PR #2, opened 2017-04-13
Fix introduced bug on rights re-writing for folders on Windows

simonferquel:
When rewriting rights for Windows, I introduced a bug removing the directory flag.
This fix that with a simple ored value.


tonistiigi:
Guess this needs to keep other high bits as well, not only `ModeDir`.

simonferquel:
You are right, I amended the PR.

PR #10, opened 2017-09-15
Add filter support for walker

tonistiigi:
Allows overriding and filtering walked files, for example for resetting uid/gid.

cc @shouze

### DIFF
--- a/walker.go
+++ b/walker.go
@@ -15,6 +15,7 @@ import (
 type WalkOpt struct {
 	IncludePaths    []string // todo: remove?
 	ExcludePatterns []string
+	Map             func(*Stat) bool
 }
 
 func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) error {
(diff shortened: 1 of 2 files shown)


walker.go
```diff
@@ -15,6 +15,7 @@ import (
 type WalkOpt struct {
 	IncludePaths    []string // todo: remove?
 	ExcludePatterns []string
+	Filter          func(*Stat) bool
```
shouze: As it’s possible to mutate the Stat struct... I’m not sure that Filter is self descriptive enough. It’s more close to map/reduce in fact.

tonistiigi: Updated

shouze:
@tonistiigi back to  can you confirm that to reset uid/gid to 0 when docker image build is in stream mode, I have to patch [buildkit sendDiffCopy](https://github.com/moby/buildkit/blob/3a18751db6d0b00c763d680795071e4654985670/session/filesync/diffcopy.go#L13) to pass a had hoc WalkOpt `Map`?

Also, can you confirm it will based in `DiffCopy` as it looks like that `TarStream` has been removed from buildkit? (so [this part](https://github.com/moby/buildkit/blob/225c57706a307b6434d46c0cfd5c61b23c18ad55/session/filesync/filesync.go#L187-L192) is currently dead code since `diffcopy` [is the only supported protocol now](https://github.com/moby/buildkit/blob/225c57706a307b6434d46c0cfd5c61b23c18ad55/session/filesync/filesync.go#L128-L134)).

PR #16, opened 2017-11-17
Don’t hang on diskwriter errors

tonistiigi:
fix for https://github.com/moby/buildkit/issues/165

@AkihiroSuda


tonistiigi:
@AkihiroSuda Yes, but it is tricky to clean up properly when we don't know what is hanging. Atm it was hanging on the `read()` and there is no real good way to unblock that other than signal the sender to cancel(but that may not work for other errors).

PR #17, opened 2017-11-21
fix fd leak on send

tonistiigi:
Before these would be cleaned up by the `gc` and could cause `too many open files` errors.

@fcrisciani


fcrisciani:
LGTM

PR #19, opened 2017-12-19
Add copy pkg

tonistiigi:
Golang replacement for https://github.com/tonistiigi/copy to support `chown`. I'll add similar one for archive as well.


copy/copy.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+		if err := os.Symlink(link, target); err != nil {
+			return errors.Wrapf(err, "failed to create symlink: %s", target)
+		}
+	case (fi.Mode() & os.ModeDevice) == os.ModeDevice:
+		if err := copyDevice(target, fi); err != nil {
+			return errors.Wrapf(err, "failed to create device")
+		}
+	default:
+		// TODO: Support pipes and sockets
```
AkihiroSuda: Is it technically feasible?

copy/copy_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	t2, err := ioutil.TempDir("", "test")
+	require.NoError(t, err)
+	defer os.RemoveAll(t2)
+
+	err = Copy(context.TODO(), filepath.Join(t1, "foo*"), t2)
+	require.Error(t, err)
+
+	err = Copy(context.TODO(), filepath.Join(t1, "foo*"), t2, AllowWildcards)
+	require.NoError(t, err, fmt.Sprintf("%+v", err))
```
AkihiroSuda: nit: extra Sprintf

copy/copy.go
tonistiigi: This is from containerd copyDir source. Pipes should definitely be possible, for sockets could just create empty files.

AkihiroSuda:
@tonistiigi is this mergeable?

tonistiigi:
@AkihiroSuda Yes, I think so. Rebased and fixed `fstest` import.

PR #20, opened 2018-01-07
receive: read stream to EOF on close

tonistiigi:
Make sure the grpc stream is fully read. Otherwise, stream completes with an error for the monitoring tools when the context is canceled.


tonistiigi:
@AkihiroSuda

PR #21, opened 2018-01-08
avoid not-exist error on walker

tonistiigi:
This can cause a race on copy when a big destination directory that is still being walked is deleted. 


tonistiigi:
@AkihiroSuda

PR #22, opened 2018-01-25
Fix an issue where syncing read only files

simonferquel:
When syncing files that are readonly (0111), the sync fails (and sometimes even panic, due to a synchronization problem).
The problem is that the file is created on the receiver with the same rights as the sender, thus making it unopenable in Write mode on the receiver.

This commit checks permission errors on opening files, record current file mode (for reapplying it on closing) and chmod to make it writeable.

Not sure how to test that properly though


AkihiroSuda:
cc @tonistiigi

PR #23, opened 2018-02-10
Implement chtimes on OpenBSD

nogoegst:


thaJeztah:
ping @tonistiigi PTAL

tonistiigi:
This could probably be in the `_nolinux.go` file with `!linux` build tag so there is no duplication with windows and darwin.

nogoegst:
@tonistiigi, it seems that `chtimes` is not implemented on darwin. :man_shrugging:

tonistiigi:
@nogoegst You mean `os.Chtimes` ? It definitely should be. https://github.com/golang/go/blob/6732fcc06df713fc737cee5c5860bad87599bc6d/src/os/file_posix.go#L5 https://github.com/golang/go/blob/6732fcc06df713fc737cee5c5860bad87599bc6d/src/os/file_posix.go#L139

So just to be clear, linux has own implementation. I think everything but linux can use `os.Chtimes` solution like suggested in here, instead of defining per-platform exceptions.

nogoegst:
Closing in favor of #27.

PR #24, opened 2018-02-22
export some functions that could be used elsewhere

jessfraz:
Updates https://github.com/jessfraz/img/issues/21

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -7,7 +7,7 @@ import (
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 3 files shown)


jessfraz:
ping @dmcgowan, I don't know if this should be here or continuity, but since I used the NewDiskWriter I think here, happy to help merge the two if you all want

jessfraz:
closing this, going to make another that just moves the copy function i needed here so they dont need to be exported

PR #30, opened 2018-05-16
Fixes an error propagation bug that resulted in a hang

tiborvass:
Signed-off-by: Tibor Vass <tibor@docker.com>

### DIFF
--- a/receive.go
+++ b/receive.go
@@ -158,6 +158,8 @@ func (r *receiver) run(ctx context.Context) error {
 			}
 
 			switch p.Type {
+			case PACKET_ERR:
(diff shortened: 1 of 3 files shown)


tonistiigi:
@tiborvass pushed an extra commit

tiborvass:
Tested in buildkit, LGTM

tonistiigi:
LGTM

tonistiigi:
@tiborvass There is a slight problem with this test that showed in CI. If the receiver isn't blocked when the walker errors, the receiver would close directly from `ctx.Done()`, without doing the roundtrip. I think this behavior is ok, but the error doesn't have the `from sender` prefix then.

PR #35, opened 2018-05-30
receive: convert file path to local form when registering.

ijc:
This matches the same transformation as is done in `DiskWriter.Handle`, which
eventually gets passed to the `AsyncDataCb` AKA `receiver.asyncDataFunc` which
performs the lookup.

This matters on Windows where `filepath.FromSlash` will convert `/` into `\`.
I've tested this (in the context of buildkit) with a Linux `sender` and Windows
`receiver`, I've no capability to test the opposite direction.

Signed-off-by: Ian Campbell <ijc@docker.com>

### DIFF
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -80,9 +80,7 @@ func (dw *DiskWriter) HandleChange(kind ChangeKind, p string, fi os.FileInfo, er
 		}
 	}()
 
-	p = filepath.FromSlash(p)
-
-	destPath := filepath.Join(dw.dest, p)
+	destPath := filepath.Join(dw.dest, filepath.FromSlash(p))
 
 	if kind == ChangeKindDelete {
 		// todo: no need to validate if diff is trusted but is it always?
--- a/receive.go
+++ b/receive.go
@@ -3,6 +3,7 @@ package fsutil
 import (
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/pkg/errors"
@@ -169,7 +170,7 @@ func (r *receiver) run(ctx context.Context) error {
 				}
 				if fileCanRequestData(os.FileMode(p.Stat.Mode)) {
 					r.mu.Lock()
-					r.files[p.Stat.Path] = i
+					r.files[filepath.FromSlash(p.Stat.Path)] = i
 					r.mu.Unlock()
 				}
 				i++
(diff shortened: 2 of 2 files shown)


tonistiigi:
I think we want to do this in other way - making sure that the path is not modified when diskwriter calls the async handle. The paths moving on the wire / in the objects should always be unix style and normalized - the conversion should only happen if some part actually wants to call syscalls.

ijc:
Ack, I'll have a look at how that can be achieved today.

ijc:
There's some subtleties wrt the expectations of the existing callback implementations. I'll just paste the commit message from the newly added commit:
```
diskwriter: pass wire/unix paths to callbacks.

The callbacks expect to operate on the wire form which is standardized on Unix
format.

For example `receiver.asyncDataFunc`, which is used as `AsyncDataCb` expects to
use the path as a key on a map lookup for an insertion done in the
`PACKET_STAT` case of `receiver.run` using the wire form.

This matters on Windows where `filepath.FromSlash` will convert `/` into `\`.

Only the `destPath` needs to be converted to the local form. I have confirmed
that all subsequent uses of `p` in `DiskWriter.Handle` are passing it to one of
the callbacks (either directly or via `DiskWriter.processChange` or
`DiskWriter.requestAsyncFileData`, that latter just calls `processChange`) as a
key or for error messages.

The callbacks reachable in this way are `NotifyCallback`, `SyncDataCb` and
`AsyncDataCb`, which are set to `recevier.notifyHashed`, `newWriteToFunc`
(ignoring `nopWriteTo`) functions and `receiver.asyncDataFunc` respectively.

The `newWriteToFunc` is adjusted to cope with this change and
`receive.asyncDataFunc` already expected a canonical form for path and does thr
correct thing.

The `receiver.notifyHashed` callback is user-supplied via
`RecieveOpt.NotifyHashed`. I have looked at the user in
`github.com/moby/buildkit/session/filesync/diffcopy.go` which uses the
`HandleChange` on a supplied `CacheUpdater`. One candidate for a supplied
object meeting this interface is
`github.com/moby/buildkit/cache/contenthash/cacheContext`. The `HandleChange`
callback there does an immediate `filepath.ToSlash` which I think will need to
be removed when buildkit is updated, I've been unable to determine if the
consequences of not doing so are more serious than an artificial cache-bust.

I've tested this (in the context of buildkit) with a Linux `sender` and Windows
`receiver`, I've no capability to test the opposite direction.

Signed-off-by: Ian Campbell <ijc@docker.com>
```

PR #36, opened 2018-05-31
Switch from x/net/context to context

thaJeztah:
ping @tonistiigi @tiborvass PTAL


AkihiroSuda:
BTW maybe we should have vendor?

PR #38, opened 2018-06-01
Fixes for non-root/symlink copy tests

### LINKED ISSUE #26: TestCopySimple fails
```
# go get github.com/tonistiigi/fsutil
# go test github.com/tonistiigi/fsutil -v -run ''


=== RUN   TestCopySimple
--- FAIL: TestCopySimple (0.01s)
	assertions.go:238: 
			Error Trace:	receive_test.go:80
			Error:      	Not equal: 
			            	expected: "sha256:2c6454be696ff984b27888e4757cfb5eb65c1a8f0d49c017596e3feefaa0c84e"
			            	actual  : "sha256:99b6ef96ee0572b5b3a4eb28f00b715d820bfd73836e59cc1565e241f4d1bb2f"
	assertions.go:238: 
			Error Trace:	receive_test.go:84
			Error:      	Not equal: 
			            	expected: "sha256:2fcd2d8fb162dbdfb2ecfe93724dc0f4e683ebff2d9f15485136958ef99ef8b7"
			            	actual  : "sha256:dd2529f7749ba45ea55de3b2e10086d6494cc45a94e57650c2882a6a14b4ff32"
	assertions.go:238: 
			Error Trace:	receive_test.go:88
			Error:      	Not equal: 
			            	expected: "sha256:51086b730b712a6c3a11651083e4eb00a095b3ef60174fb8934542e8cebcecca"
			            	actual  : "sha256:eca07e8f2d09bd574ea2496312e6de1685ef15b8e6a49a534ed9e722bcac8adc"
	assertions.go:238: 
			Error Trace:	receive_test.go:137
			Error:      	Not equal: 
			            	expected: "sha256:51086b730b712a6c3a11651083e4eb00a095b3ef60174fb8934542e8cebcecca"
			            	actual  : "sha256:eca07e8f2d09bd574ea2496312e6de1685ef15b8e6a49a534ed9e722bcac8adc"
	assertions.go:238: 
			Error Trace:	receive_test.go:141
			Error:      	Not equal: 
			            	expected: "sha256:f2b015030e257b821d76681d822e23af06de1621e3b1a9168fbc9d22e83ed345"
			            	actual  : "sha256:cd14a931fc2e123ded338093f2864b173eecdee578bba6ec24d0724272326c3a"
```

Test with both go 1.7 and go 1.10, in case it matters.

tonistiigi:
fixes #26 

@thaJeztah Thanks for bringing my attention to this.

There are some fixes in here that are mostly just for making the test compatible with non-linux and non-root user, but two cases also need to be checked when vendoring this to buildkit. These are:

- handling the size on directory records (now resets to 0)
- handling mode of symlinks. This is tricky because it is only supported on linux (and even then with some crazy hacks). I think we need to detect it in buildkit hash function as well and force set.

This also fixes the chtimes issue that showed up on darwin. I'm not sure why this one was not failing for anyone else.

### DIFF
--- a/walker.go
+++ b/walker.go
@@ -131,13 +131,13 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 		stat := &Stat{
 			Path:    path,
 			Mode:    uint32(fi.Mode()),
-			Size_:   fi.Size(),
 			ModTime: fi.ModTime().UnixNano(),
 		}
 
 		setUnixOpt(fi, stat, path, seenFiles)
 
 		if !fi.IsDir() {
+			stat.Size_ = fi.Size()
 			if fi.Mode()&os.ModeSymlink != 0 {
 				link, err := os.Readlink(origpath)
 				if err != nil {
--- a/chtimes_nolinux.go
+++ b/chtimes_nolinux.go
@@ -9,5 +9,12 @@ import (
 
 func chtimes(path string, un int64) error {
 	mtime := time.Unix(0, un)
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return err
+	}
+	if fi.Mode()&os.ModeSymlink != 0 {
+		return nil
+	}
 	return os.Chtimes(path, mtime, mtime)
 }
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -102,8 +102,10 @@ func (dw *DiskWriter) HandleChange(kind ChangeKind, p string, fi os.FileInfo, er
 		return errors.Errorf("%s invalid change without stat information", p)
 	}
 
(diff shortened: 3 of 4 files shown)


walker.go
```diff
@@ -18,7 +18,7 @@ type WalkOpt struct {
 	Map             func(*Stat) bool
 }
 
-func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) error {
+func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) (err error) {
```
thaJeztah: Perhaps good to rename the output variable to prevent confusion

tonistiigi: updated

PR #41, opened 2018-06-07
copy: do not chmod destination dir if it exists

tiborvass:
Signed-off-by: Tibor Vass <tibor@docker.com>


thaJeztah:
Is there a test that needs to be updated for this?

tiborvass:
@thaJeztah you're too fast (or I'm too slow!). Updated with a test for both this case, and for #33

PR #55, opened 2018-10-30
copy: add XAttrErrorHandler

AkihiroSuda:
@tonistiigi


tonistiigi:
Do we need a handler function? Maybe just a boolean would be enough.

AkihiroSuda:
For allowing logging https://github.com/tonistiigi/copy/pull/16/files#diff-d606cbe99da719b67266e4e985aee64fR138

PR #57, opened 2019-01-30
update go.sum with go1.11.5

tonistiigi:
It seems that there was a bugfix between `go1.11` and `go1.11.5` that causes `go.sum` checksums to differ and `go mod tidy` to fail on `go1.11.5` with: 

```
verifying gotest.tools@v2.2.0+incompatible: checksum mismatch
    downloaded: h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
    go.sum:     h1:y0IMTfclpMdsdIbr6uwmJn5/WZ7vFuObxDMdrylFM3A=
```

New go.sum generated with go1.11.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
 language: go
 
 go: 
- - "1.11"
+ - "1.11.x"
 
 
 install:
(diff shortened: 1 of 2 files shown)


tonistiigi:
So the new hashes do not even build with `go1.11` that travis has. DO NOT merge

tonistiigi:
This breaks people who have not upgraded to 1.11.4/5 so that they can't even build the project. I think its better to just let the `tidy` fail. With the old invalid hashes it still seems to build fine with `1.11.5`

tonistiigi:
Never mind. `go build` does not work the other way as well (while verify always work). 😞

PR #60, opened 2019-03-11
copy: updates for buildkit fileop

tonistiigi:
updates for the copy package required for https://github.com/moby/buildkit/pull/809

Contains fs root protection as the code doesn't need to run in a container anymore, and support for non-followlinks mode. Some of the previous functionality is refactored to exposed more public functions for controlling wildcards and mkdirp.

@AkihiroSuda @tiborvass

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -21,7 +21,23 @@ var bufferPool = &sync.Pool{
 	},
 }
 
-func ResolveWildcards(root, src string) ([]string, error) {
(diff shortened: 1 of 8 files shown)


tonistiigi:
@AkihiroSuda Merging to include it in the vendoring I need to do anyway. Comment if you need anything changed for moby/buildkit#809 and I'll fix in follow-up.

PR #65, opened 2019-03-27
add tarwriter

tonistiigi:
Add tarwriter and refactor `FS` so that memory-only subdirs can be exported with the tarball.

@AkihiroSuda

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -3,9 +3,11 @@ package fsutil
 import (
 	"context"
 	"io"
+	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


fs.go
```diff
@@ -37,36 +39,80 @@ func (fs *fs) Open(p string) (io.ReadCloser, error) {
...
 }
 
 func (fs *subDirFS) Walk(ctx context.Context, fn filepath.WalkFunc) error {
-	main := &StatInfo{Stat: &fs.stat}
-	if !main.IsDir() {
-		return errors.Errorf("fs subdir not mode directory")
+	for _, d := range fs.dirs {
+		fi := &StatInfo{Stat: &d.Stat}
+		if !fi.IsDir() {
+			return errors.Errorf("fs subdir not mode directory")
```
AkihiroSuda: errorf d.Stat.Path?

fs.go
```diff
@@ -37,36 +39,80 @@ func (fs *fs) Open(p string) (io.ReadCloser, error) {
...
+		}
+		if err := d.FS.Walk(ctx, func(p string, fi os.FileInfo, err error) error {
+			stat, ok := fi.Sys().(*types.Stat)
+			if !ok {
+				return errors.Wrapf(err, "invalid fileinfo without stat info: %s", p)
+			}
+			stat.Path = path.Join(d.Stat.Path, stat.Path)
+			if stat.Linkname != "" {
+				if fi.Mode()&os.ModeSymlink != 0 {
+					if strings.HasPrefix(stat.Linkname, "/") {
```
AkihiroSuda: can we move this to `_unix.go`?

tonistiigi: What part specifically are you concerned about? `types.Stat` is supposed to be cross platform and always normalized to unix slashes. (unlike syscall.Stat_t)

fs.go
tonistiigi: updated

PR #66, opened 2019-04-03
copy: add fast copy path for darwin

tonistiigi:

### DIFF
--- a/copy/copy_linux.go
+++ b/copy/copy_linux.go
@@ -52,6 +52,21 @@ func (c *copier) copyFileInfo(fi os.FileInfo, name string) error {
 	return nil
 }
 
+func copyFile(source, target string) error {
+	src, err := os.Open(source)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


tiborvass:
Should you have an `_other.go` file with `copyFile` to avoid compile errors on other platforms for code importing this package?

tonistiigi:
It already exists for linux/windows. We don't have any way to test others anyway.

copy/copy_linux.go
```diff
@@ -52,6 +52,21 @@ func (c *copier) copyFileInfo(fi os.FileInfo, name string) error {
 	return nil
 }
 
+func copyFile(source, target string) error {
```
tiborvass: you could call this `copyFileNaive` in `copy.go` and call it as a fallback from darwin.

tonistiigi: But this function is not naive in linux. It does copy_file_range on linux.

PR #69, opened 2019-10-08
receive: use filter on receive diff

tonistiigi:
Make sure filter is taken into an account when comparing existing files in the receiver. This issue prevents incremental transfers with userns because the metadata on disk is always different than the one being transferred.

### DIFF
--- a/diff_containerd.go
+++ b/diff_containerd.go
@@ -87,7 +87,15 @@ func doubleWalkDiff(ctx context.Context, changeFn ChangeFunc, a, b walkerFn, fil
 			}
 
 			var f *types.Stat
-			k, p := pathChange(f1, f2)
+			var f2copy *currentPath
+			if f2 != nil {
+				statCopy := *f2.stat
+				if filter != nil {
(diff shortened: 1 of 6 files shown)


diff_containerd.go
```diff
@@ -86,45 +86,53 @@ func doubleWalkDiff(ctx context.Context, changeFn ChangeFunc, a, b walkerFn) (er
...
 					return err
 				}
-				if f1.f.IsDir() && !f2.f.IsDir() {
+				if f1.stat.IsDir() && !f2copy.stat.IsDir() {
 					rmdir = f1.path + string(os.PathSeparator)
 				} else if rmdir != "" {
 					rmdir = ""
 				}
-				f = f2.f
+				f = f2.stat
```
tiborvass: @tonistiigi can you explain why sometimes the original `f2` is used and at other times the `f2copy` copy is used?

tonistiigi: because sometimes we need the filtered and sometimes not. checks are with filter but we pass the original to the callback.

PR #71, opened 2020-01-25
Create local atime/mtime wrappers for syscall.Stat_t

TBBle:
These wrappers replace the use of the utility functions in `github.com/containerd/containerd/sys`, and work around https://github.com/golang/go/issues/32887.

The implementations were derived from https://golang.org/search?q=Stat_t#Global_pkg/syscall

Hence, `containerd/containerd` is no longer used and can be removed as a dependency, which also removes `Microsoft/hcsshim`.

### DIFF
--- a/copy/copy_linux.go
+++ b/copy/copy_linux.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"syscall"
 
-	"github.com/containerd/containerd/sys"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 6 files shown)


TBBle:
The motivation for this change is that it seems this library is causing go to vendor an old version of hcsshim into buildkit, even if a new version of containerd (which has had newer hcsshim vendored in) is vendored into buildkit.

As far as I can tell, nothing except containerd has any business trying to pull hcsshim into buildkit...

tonistiigi:
@TBBle Could you rebase so the CI runs in other platforms. Thanks.

PR #74, opened 2020-02-25
reran `go generate`

AkihiroSuda:
For https://github.com/moby/buildkit/pull/1379


types/stat.pb.go
```diff
@@ -35,7 +24,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.GoGoProtoPackageIsVersion2 // please upgrade the proto package
+const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
```
AkihiroSuda: Maybe we want to keep v2, but not sue how to do so.

PR #82, opened 2020-05-12
Drop support for go1.12, and use errors.Is() for error-detection

thaJeztah:
follow-up to https://github.com/tonistiigi/fsutil/pull/81

### DIFF
--- a/walker.go
+++ b/walker.go
@@ -219,12 +219,5 @@ func trimUntilIndex(str, sep string, count int) string {
 }
 
 func isNotExist(err error) bool {
-	err = errors.Cause(err)
-	if os.IsNotExist(err) {
-		return true
-	}
-	if pe, ok := err.(*os.PathError); ok {
-		err = pe.Err
-	}
-	return err == syscall.ENOTDIR
+	return errors.Is(err, os.ErrNotExist) || errors.Is(err, syscall.ENOTDIR)
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ services:
 language: go
 
 go:
-  - "1.12.x"
   - "1.13.x"
   - "1.14.x"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,10 @@ services:
 
 language: go
 
-go: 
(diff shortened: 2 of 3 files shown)


walker.go
```diff
@@ -219,10 +219,10 @@ func trimUntilIndex(str, sep string, count int) string {
...
 func isNotExist(err error) bool {
-	err = errors.Cause(err)
-	if os.IsNotExist(err) {
+	if errors.Is(err, os.ErrNotExist) || errors.Is(err, syscall.ENOTDIR) {
 		return true
 	}
+	err = errors.Cause(err)
 	if pe, ok := err.(*os.PathError); ok {
 		err = pe.Err
 	}
```
thaJeztah, line 228: Wondering if this code is still needed,  or if the `errors.Is(err, syscall.ENOTDIR)` I added above captures this already

@AkihiroSuda @tonistiigi ?

thaJeztah:
Oh! This doesn't work because travis still tests against go 1.12 as well

thaJeztah:
```
The command "sudo ./test -test.v" exited with 0.
7.71s$ GOOS=darwin go build ./...
# github.com/tonistiigi/fsutil
./stat_unix.go:17:6: undefined: "github.com/pkg/errors".Is
./walker.go:222:5: undefined: "github.com/pkg/errors".Is
note: module requires Go 1.13
```

thaJeztah:
I removed go1.12, and added go1.13, go1.14

If we want to continue support for go < 1.13, I can add `_go112` and `_go113` files to use the old implementation for older Go versions

walker.go
tonistiigi: This should use `errors.As` instead of `Cause`

tonistiigi: Didn't see the line comparing to `syscall.ENOTDIR` below. I guess this part isn't needed at all then now.

thaJeztah: Thanks! Yes, I was in doubt if `errors.Is()` would do everything that type-matching `err.(*os.PathError)` did. I'll remove the part

thaJeztah:
Rebased; should be ready to go

PR #88, opened 2020-07-24
Add github action CI

tonistiigi:
Add CI based on github actions through bake targets. Travis can remain enabled for now as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,45 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - master
+      - gh_test_ci
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        GO_VERSION: ["1.13", "1.14"]
+    env:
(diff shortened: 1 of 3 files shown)


thaJeztah:
- Do we want these to run on forks as well? Otherwise this trick may be used https://github.com/docker/docker.github.io/commit/f2e1c94e9e56b3a3c634cc41af5b9cd5ada23438#diff-01777e4a9846fea5f3fcc8fe40d44680
- Should we set `timeout-minutes` ? (The default is 6 hours IIRC)


/cc @stefanscherer who has done more with GH Actions

StefanScherer:
Yes, we should always set `timeout-minutes` like here https://github.com/docker/docker.github.io/blob/master/.github/workflows/build-master.yml#L12 
Otherwise LGTM

StefanScherer:
As long as there are no secrets in the pipeline it's fine to let people fork the repo + pipeline to test in their repo first.

tonistiigi:
> Do we want these to run on forks as well? Otherwise this trick may be used

I don't see why not. If someone forks the repository there should be no need for them to modify the yaml to run CI.

PR #101, opened 2021-04-21
Add IncludePatterns and ExcludePatterns options for Copy

aaronlehmann:
Allow include patterns and exclude patterns to be specified, similarly
to `Walker`.

There is a bit of extra complexity to handle the case of a pattern like
`a/*/c`. In this case, creating the directories `a` and `a/b` may need
to be delayed until we encounter `a/b/c`.

cc @hinshun

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -11,7 +11,9 @@ import (
 	"time"
 
 	"github.com/containerd/continuity/fs"
+	"github.com/docker/docker/pkg/fileutils"
 	"github.com/pkg/errors"
+	"github.com/tonistiigi/fsutil/prefix"
 )
 
 var bufferPool = &sync.Pool{
@@ -86,7 +88,10 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		return err
 	}
 
-	c := newCopier(ci.Chown, ci.Utime, ci.Mode, ci.XAttrErrorHandler)
+	c, err := newCopier(ci.Chown, ci.Utime, ci.Mode, ci.XAttrErrorHandler, ci.IncludePatterns, ci.ExcludePatterns)
+	if err != nil {
+		return err
+	}
 	srcs := []string{src}
 
 	if ci.AllowWildcards {
@@ -109,7 +114,8 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		if err != nil {
 			return err
 		}
-		if err := c.copy(ctx, srcFollowed, dst, false); err != nil {
+		includeAll := len(c.includePatterns) == 0
+		if err := c.copy(ctx, srcFollowed, "", dst, false, includeAll); err != nil {
 			return err
 		}
 	}
@@ -162,6 +168,10 @@ type CopyInfo struct {
 	XAttrErrorHandler XAttrErrorHandler
 	CopyDirContents   bool
 	FollowLinks       bool
+	// Include only files/dirs matching at least one of these patterns
+	IncludePatterns []string
+	// Exclude files/dir matching any of these patterns (even if they match an include pattern)
+	ExcludePatterns []string
 }
 
 type Opt func(*CopyInfo)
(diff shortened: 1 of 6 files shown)


copy/copy.go
```diff
@@ -197,48 +207,110 @@ func AllowXAttrErrors(ci *CopyInfo) {
...
+			return err
+		}
+		if exclude {
+			return nil
+		}
+	}
+
 	if !fi.IsDir() {
+		if include {
+			if err := c.createParentDirs(src, srcComponents, target, overwriteTargetMetadata); err != nil {
```
tonistiigi, line 294: Isn't this quite a significant added performance overhead to rerun these checks where most of the time directories exist? If true then I think we need some caching. As everything is sorted shouldn't take too much memory.

copy/copy.go
```diff
@@ -197,48 +207,110 @@ func AllowXAttrErrors(ci *CopyInfo) {
...
 
 	copyFileInfo := true
 
 	switch {
 	case fi.IsDir():
-		if created, err := c.copyDirectory(ctx, src, target, fi, overwriteTargetMetadata); err != nil {
+		if created, err := c.copyDirectory(ctx, src, srcComponents, target, fi, overwriteTargetMetadata, includeAll); err != nil {
 			return err
-		} else if !overwriteTargetMetadata {
+		} else if !overwriteTargetMetadata || !includeAll {
```
tonistiigi, line 313: bit confused about the includeAll check in here as exclude patterns seem to be not affected by this.

copy/copy.go
```diff
@@ -282,26 +354,127 @@ func (c *copier) copy(ctx context.Context, src, target string, overwriteTargetMe
...
-func (c *copier) copyDirectory(ctx context.Context, src, dst string, stat os.FileInfo, overwriteTargetMetadata bool) (bool, error) {
+func (c *copier) include(path string, fi os.FileInfo) (bool, bool, error) {
+	matched := false
+	partial := true
+	for _, pattern := range c.includePatterns {
+		if fi.IsDir() {
+			pattern = strings.TrimSuffix(pattern, string(filepath.Separator))
+		}
+
+		if ok, p := prefix.Match(pattern, path); ok {
```
tonistiigi, line 365: Why do IncludePatterns/ExcludePatterns use different format/algorithm? Afaics only `ExcludePatterns` support `!`. Is there a difference of `ExcludePatterns` and potentially doing `!` in `IncludePatterns`.

copy/copy.go
```diff
@@ -282,26 +354,127 @@ func (c *copier) copy(ctx context.Context, src, target string, overwriteTargetMe
...
+	if c.excludePatternMatcher == nil {
+		return false, nil
+	}
+
+	m, err := c.excludePatternMatcher.Matches(path)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to match excludepatterns")
+	}
+	if m {
+		if fi.IsDir() && c.excludePatternMatcher.Exclusions() {
```
tonistiigi, line 393: Not for this PR but this looks non-optimal. I guess `!` is rarely used in practice.

copy/copy.go
```diff
@@ -109,7 +114,8 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		if err != nil {
 			return err
 		}
-		if err := c.copy(ctx, srcFollowed, dst, false); err != nil {
+		includeAll := len(c.includePatterns) == 0
```
tonistiigi, line 117: Excludepatterns don't matter in here?

aaronlehmann: `includeAll` means we can skip evaluating the include patterns. Usually it's used when a parent dir matches an include pattern. It doesn't skip evaluating exclude patterns.

Maybe `skipIncludePatterns` would be a better name?

copy/copy.go
aaronlehmann: This does add overhead, but only to the case where an include pattern is matched - so it won't regress performance of any existing cases.

Happy to add caching if you think it's worth the complexity.

copy/copy_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/containerd/continuity/fs/fstest"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
aaronlehmann: Not sure I follow, do you prefer `require` over `assert`? I generally use `assert` unless the failed assertion would put the test in a bad state where it doesn't make sense to continue.

prefix/match.go
```diff
@@ -0,0 +1,32 @@
+package prefix
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+func Match(pattern, name string) (bool, bool) {
+	count := strings.Count(name, string(filepath.Separator))
```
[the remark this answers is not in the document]
aaronlehmann: I think you are right about this. I guess https://github.com/moby/buildkit/pull/2082/files#diff-364fe775b55b3eec6b3ea0859ab920f0981c61d377e47830a46de5ec04058c44R437 does not fail on Windows because there are no test cases that depend on parent dirs being included in the checksum. Not sure if there's a way to simulate a change in parent dir metadata in that test so we can capture that scenario.

copy/copy.go
aaronlehmann: Modeled on this code from `walker.go`: https://github.com/tonistiigi/fsutil/blob/8599091d69851a37058eec95da25c794ac060a0a/walker.go#L131

copy/copy.go
aaronlehmann: `created` will always be false if `includeAll` is false (because in that case we delay creation of parent dirs). So this is shorter version of:

```go
if !includeAll {
        // directory may not have been created yet, so don't try to set its metadata
        copyFileInfo = false
}
```

copy/copy.go
aaronlehmann: I wanted to match the logic in https://github.com/tonistiigi/fsutil/blob/8599091d69851a37058eec95da25c794ac060a0a/walker.go#L87

I think I'm following the same logic (it has separate handling for `IncludePatterns` and `ExcludePatterns` and doesn't appear to handle `!` in `IncludePatterns`), but maybe I'm missing something.

copy/copy_test.go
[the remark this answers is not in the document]
tonistiigi: We usually use `require` exclusively so it is a bit out of place and should have a comment so that it is not reverted in the future for consistency.

copy/copy.go
tonistiigi: I think this is needed. The number of syscalls per file is quite important for copy.

copy/copy.go
tonistiigi: I see. I don't remember why it is like this. I guess for `.dockerignore` is the only use-case atm that required `!` and that is the excludes list. `IncludePatterns` is not used at all anymore as well and `FollowPaths` is used instead. I guess the `FollowPaths` case where symlinks need to be followed does not apply here?

I think it is important that something like `COPY --filter` can be implemented with this. I'm not quite sure if atm client would need to convert some of the patterns to a negative version. But I guess if we would only use one input then with current implementation depending on if user wants to mostly define include or exclude paths one of them would be much slower. Matching with the options of walker is not the most important here so if you think there is a better/simpler logic we can use instead lmk.

copy/copy.go
```diff
@@ -282,43 +361,157 @@ func (c *copier) copy(ctx context.Context, src, target string, overwriteTargetMe
...
-func (c *copier) copyDirectory(ctx context.Context, src, dst string, stat os.FileInfo, overwriteTargetMetadata bool) (bool, error) {
+func (c *copier) include(path string, fi os.FileInfo) (bool, bool, error) {
+	matched := false
+	partial := true
+	for _, pattern := range c.includePatterns {
+		if fi.IsDir() {
+			pattern = strings.TrimSuffix(pattern, string(filepath.Separator))
+		}
+
+		if ok, p := prefix.Match(pattern, path, false); ok {
```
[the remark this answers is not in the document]
aaronlehmann: That could potentially overwrite a `true` value of `matched` with `false`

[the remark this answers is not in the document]
hinshun: Ah missed that.

aaronlehmann:
@tonistiigi: Any more feedback on this PR?

tonistiigi (approved):
As commented earlier, different types for include/exclude is a bit of a problem but don't have a specific alternative proposal atm.

aaronlehmann:
Thanks!

PR #102, opened 2021-05-03
Fix compilation on FreeBSD

gizahNL:
Currently fsutil is not building on FreeBSD, these changes make it build (again?)


tonistiigi (changes requested):
PTAL the CI failures. Also, add freebsd to the cross target so it is verified in CI.

gizahNL:
> PTAL the CI failures. Also, add freebsd to the cross target so it is verified in CI.

I'll fix it up tonight :)

copy/copy_freebsdsolaris.go
```diff
@@ -0,0 +1,33 @@
+//+build freebsd solaris
+
+package fs
+
+import (
+	"github.com/pkg/errors"
+	"os"
+	"io"
+)
```
thaJeztah, line 9: nit: `github.com/pkg/errors` should be in a separate "group" here (not a blocker)

```suggestion
import (
	"os"
	"io"

	"github.com/pkg/errors"
)
```

thaJeztah (commented):
looks like CI is green now; I left one nit (also probably needs commits to be squashed; at least the second commit)

thaJeztah:
@tonistiigi ptal

copy/copy_freebsdsolaris.go
tonistiigi: Are we missing some ci validation for this?

thaJeztah: Hmm.. perhaps, or the linter doesn't catch it (as it's "optional" to group them)

tonistiigi: The formatter or goimports should catch it

gizahNL:
ptal

PR #103, opened 2021-05-24
Use PatternMatcher for copy include patterns as well as exclude patterns

aaronlehmann:
This changes the logic for copy's "include patterns" to use
`PatternMatcher`, so it behaves the same as "exclude patterns". This is a
net simplification and provides more consistency between the two
parameters (for example, include patterns can now specify negative
matches), but is a bit worse performance-wise (we can't skip a directory
which fails to partially match a pattern, because files inside it may
match the pattern).

cc @hinshun

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -321,7 +321,7 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 
 	switch {
 	case fi.IsDir():
-		if created, err := c.copyDirectory(ctx, src, srcComponents, target, fi, overwriteTargetMetadata, skipIncludePatterns); err != nil {
+		if created, err := c.copyDirectory(ctx, src, srcComponents, target, fi, overwriteTargetMetadata, skipIncludePatterns, include); err != nil {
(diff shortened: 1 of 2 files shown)


aaronlehmann:
One concern I have is that `PatternMatcher` seems to make assumptions that it is only used for exclude patterns, such as these messages:

```go
				return nil, errors.New("illegal exclusion pattern: \"!\"")
```

```go
	if matched {
		logrus.Debugf("Skipping excluded path: %s", file)
	}
```

If we move forward with this, we might want to create a separate PR against `docker/docker` to adjust these.

tonistiigi (commented):
Could we have test cases that check for `**` and include and exclude patterns together.

tonistiigi:
> but is a bit worse performance-wise (we can't skip a directory
which fails to partially match a pattern, because files inside it may
match the pattern)

I guess if we wanted to we could have a follow-up that detects if the patterns can be optimized or if there are ones where slow path is needed.

tonistiigi (approved):

PR #108, opened 2021-08-17
Simplify walker to use PatternMatcher instead of custom matching

aaronlehmann:
This changes the pattern matching implementation in walker to use `PatternMatcher`, making it consistent with the "copy" code.

This makes it support `**` patterns as well, which formerly didn't work.

The added benchmark shows the new implementation is slightly faster than the old one.

Opened on top of #107 - the first commit is from that PR, so only look at the second one.

Replaces #79

cc @coryb


aaronlehmann:
Rebased now that #107 is merged.

walker_test.go
```diff
@@ -305,3 +304,361 @@ func tmpDir(inp []*change) (dir string, retErr error) {
...
+	trimEqual(t, `
+		dir a
+		dir a/b
+		dir a/b/bar
+		file a/b/bar/foo
+		dir bar
+		file bar/foo
+	`, string(b.Bytes()))
+
+	// TODO can we support this? Currently filepath.Clean is called
```
coryb, line 568: I think we can replace this comment/block with:
```
	b.Reset()
	err = Walk(context.Background(), d, &WalkOpt{
		IncludePatterns: []string{"**/foo/**"},
	}, bufWalk(b))
	assert.NoError(t, err)

	trimEqual(t, `
		dir foo
		dir foo/bar
		file foo/bar/bee
	`, string(b.Bytes()))
```

I had originally thought `**/foo/` could match all directories named `foo`, but I think `**/foo/**` better represents this idea and seems to work.

coryb (approved):
Looks good to me, thanks for carrying this work!

walker_test.go
aaronlehmann: Updated, thanks.

aaronlehmann:
@tonistiigi: Do you see any compatibility concerns with this change?

tonistiigi (approved):

PR #111, opened 2021-10-13
ci: test against macos

crazy-max:
Follow-up #110

Keep the same go matrix. Feel free to change it when merged.

cc @tonistiigi 

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>


crazy-max:
As expected darwin tests fail:

image

Still waiting for Windows runners to start (might be an issue with the infra).

https://github.com/crazy-max/fsutil/runs/3879564813?check_suite_focus=true

PR #113, opened 2021-10-14
copy: log changes made by copy

tonistiigi:
This allows capturing changes made by copy so that buildkit diff can ignore the parent directories when layer blobs are created.

@sipsma

Was wondering if all parent should be ignored or only when parent directory existed. Went with the former for now. In buildkit that would mean that in some cases (eg. `--chown/--chmod`) parent dir needs to be manually created before copy. Same question about includepatterns that is little more tricky as parents of the patterns can't be predicted outside of the copy.


sipsma (approved):
> Was wondering if all parent should be ignored or only when parent directory existed. Went with the former for now

That makes sense to me and seems to fit what we want for the cases you brought up in the other issue.

PR #120, opened 2021-12-21
Handle EPERM errors when walking excluded dirs

aaronlehmann:
A recent change to `Walk` causes it to walk dirs that matched an exclude
pattern, in cases where there are exceptions to the exclude pattern (do
we have better terminology for this?). This was done in order to support
exclusions like:

    foo
    !**/bar

In that example, we want to walk `foo` even though it's excluded,
because there might be a file or directory `bar` somewhere underneath
it.

However, this change in behavior surfaced "permission denied" errors
that didn't show up before. In the case that there is no permission to
list contents of "foo", that now results in an error.

To handle this properly, we need to ignore EPERM errors on dirs that we
are only accessing to search for these hypothetical nested files/dirs
which might end up included.

A followup change will optimize this behavior by avoiding the search in
cases where such exclusions can't possibly apply based on the patterns
involved.

### DIFF
--- a/walker.go
+++ b/walker.go
@@ -83,15 +83,12 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 	var parentDirs []visitedDir
 
 	seenFiles := make(map[uint64]string)
-	return filepath.Walk(root, func(path string, fi os.FileInfo, err error) (retErr error) {
+	return filepath.Walk(root, func(path string, fi os.FileInfo, walkErr error) (retErr error) {
 		defer func() {
 			if retErr != nil && isNotExist(retErr) {
 				retErr = filepath.SkipDir
 			}
 		}()
-		if err != nil {
-			return err
-		}
 
 		origpath := path
 		path, err = filepath.Rel(root, path)
@@ -165,6 +162,13 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 				}
 				skip = true
 			}
+		}
+
+		if walkErr != nil {
+			if skip && errors.Is(walkErr, os.ErrPermission) {
+				return nil
+			}
+			return walkErr
 		}
 
 		if includeMatcher != nil || excludeMatcher != nil {
--- a/walker_test.go
+++ b/walker_test.go
@@ -8,11 +8,13 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/tonistiigi/fsutil/types"
 )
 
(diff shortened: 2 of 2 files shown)


aaronlehmann:
The `test-macos 1.16` failure looks random, since #121 includes this commit and passed.

aaronlehmann:
Passed on a second attempt.

aaronlehmann:
@tonistiigi: Do you want to move forward with this?

walker.go
```diff
@@ -167,6 +164,13 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 			}
 		}
 
+		if walkErr != nil {
+			if skip && errors.Is(walkErr, os.ErrPermission) {
```
tonistiigi, line 168: with this `skip` check in here is it possible to ever reach this code when #121 is applied?

aaronlehmann: Yes, the optimization in #121 only applies in specific cases when only simple patterns are used and we can easily prove that a pattern won't match anything inside a dir. We will still set `skip = true` when we can't prove this and need to test the files inside the dir. We would reach this code when we recurse into a directory in these cases.

tonistiigi (approved):

tonistiigi:
@aaronlehmann 

Looks like this has caused a panic: 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xdf92d2]

goroutine 351 [running]:
github.com/tonistiigi/fsutil.Walk.func1({0xc000878140, 0xe}, {0x0, 0x0}, {0x213e7e0, 0xc0008d8120})
        /src/vendor/github.com/tonistiigi/fsutil/walker.go:137 +0xfb2
path/filepath.walk({0x1e72730, 0x1}, {0x219a120, 0xc0005efa00}, 0xc000431d48)
        /usr/local/go/src/path/filepath/path.go:438 +0x20d
path/filepath.Walk({0x1e72730, 0x1}, 0xc0005c0d48)
        /usr/local/go/src/path/filepath/path.go:505 +0x6c
github.com/tonistiigi/fsutil.Walk({0x217c700, 0xc000a1da00}, {0x7fff8b90190e, 0x1}, 0xc00057d8b0, 0xc0005372f0)
        /src/vendor/github.com/tonistiigi/fsutil/walker.go:109 +0x95d
github.com/tonistiigi/fsutil.(*fs).Walk(0xc0007f2800, {0x217c700, 0xc000a1da00}, 0xc00014399b)
        /src/vendor/github.com/tonistiigi/fsutil/fs.go:36 +0x3d
github.com/tonistiigi/fsutil.(*sender).walk(0xc0009c1320, {0x217c700, 0xc000a1da00})
        /src/vendor/github.com/tonistiigi/fsutil/send.go:147 +0xa6
github.com/tonistiigi/fsutil.(*sender).run.func1()
        /src/vendor/github.com/tonistiigi/fsutil/send.go:60 +0x2e
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /src/vendor/golang.org/x/sync/errgroup/errgroup.go:57 +0x67
created by golang.org/x/sync/errgroup.(*Group).Go
        /src/vendor/golang.org/x/sync/errgroup/errgroup.go:54 +0x92
```

With the removal of the early error check inside the walker iteration, `fi` can now be accessed before the error check is reached, causing a `nil dereference`. Not sure why the tests didn't catch this. Is it because the code paths that call `fi.IsDir` are not reached by the tests or because `fi` is nil only on certain errors.

PR #130, opened 2022-05-02
walker: add an API so that Map-based walks can skip directories.

nicks:
Related to feature request https://github.com/moby/buildkit/issues/2837


nicks:
I debated what the exact function name should be here; happy to go with something different.

tonistiigi:
https://github.com/moby/buildkit/issues/2837#issuecomment-1112770883 

> Maybe instead of adding a new method, making the old one return an enum type is better.

wdyt? This would be more in line with the stdlib `SkipDir` feature.

nicks:
sure! if we don't mind breaking backwards compatibility with the API. Here's a version that returns an enum.

walker.go
```diff
@@ -19,9 +19,29 @@ type WalkOpt struct {
 	// FollowPaths contains symlinks that are resolved into include patterns
 	// before performing the fs walk
 	FollowPaths []string
-	Map         FilterFunc
+	Map         WalkFunc
 }
 
+type WalkFunc func(string, *types.Stat) WalkResult
```
tonistiigi, line 25: I guess this is more like a `FilterFunc/FilterResult`

nicks: ya, the problem is that there's already a FilterFunc
https://github.com/tonistiigi/fsutil/blob/master/diskwriter.go#L30
and the places it's used, Walk/Skip doesn't really make sense

maybe we could call it WalkFilterFunc / WalkFilterResult?

tonistiigi: Maybe just `MapFunc` then

nicks: :+1: , updated!

tonistiigi (commented):
LGTM but squash the commits

nicks:
squashed!

tonistiigi (approved):

PR #132, opened 2022-05-05
copy: don't copy file info of non-existent dir.

sipsma:
Before this, if a directory was excluded, copy would still end up
getting called with overwriteTargetMetadata=true in the code path where
subdirs of the root of the src are iterated over. This meant that even
though the target directory was never created and never existed, the
copyFileInfo bool would stay its default value of true and the code
would try to copy file info to a directory that doesn't exist.

This update just adds a check to see if the target dir actually exists
and sets copyFileInfo to false if not.

Signed-off-by: Erik Sipsma <erik@sipsma.dev>

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -335,7 +335,7 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 		}
 	}
 
-	copyFileInfo := true
+	copyFileInfo := include
 	notify := true
 
 	switch {
@@ -345,7 +345,9 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 			include, includeMatchInfo, excludeMatchInfo,
 		); err != nil {
 			return err
-		} else if !overwriteTargetMetadata || c.includePatternMatcher != nil {
+		} else if !overwriteTargetMetadata {
+			// if we aren't supposed to overwrite existing target metadata,
+			// then we only need to copy file info if we newly created it
 			copyFileInfo = created
 		}
 		notify = false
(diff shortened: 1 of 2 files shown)


sipsma:
cc @tonistiigi if this is indeed a bug then it was hit IRL here: https://github.com/dagger/dagger/issues/2401

copy/copy.go
```diff
@@ -340,14 +340,26 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
...
+		created, err := c.copyDirectory(
 			ctx, src, srcComponents, target, fi, overwriteTargetMetadata,
 			include, includeMatchInfo, excludeMatchInfo,
-		); err != nil {
+		)
+		if err != nil {
 			return err
-		} else if !overwriteTargetMetadata || c.includePatternMatcher != nil {
+		}
+		if !overwriteTargetMetadata || c.includePatternMatcher != nil {
```
aaronlehmann, line 350: The `|| c.includePatternMatcher != nil` here looks bogus. I think that should be removed.

sipsma:
@aaronlehmann 

> I wonder if it would be more correct to only copy file info when include == true. Otherwise, we'd be applying file info to a directory that we aren't intentionally copying, but happens to incidentally exist.

Agree with this. Interestingly, when I reverted my commit and then only applied your other suggestion to remove `|| c.includePatternMatcher != nil` a lot of existing test cases started failing.

But when I then also changed `copyFileInfo` to default to `include` instead of `true`, everything passed again.

I can't say I currently understand why the `|| c.includePatternMatcher != nil` was load-bearing before but it does seem like whatever it was it is also handled by just defaulting `copyFileInfo` to `include`, at least according to the tests.

I've updated my commit to do this for now. Let me know if there's any concerns.

aaronlehmann:
LGTM

sipsma:
@tonistiigi ping

tonistiigi (approved):

PR #135, opened 2022-06-06
Receive multiple dests

a-palchikov:
This PR implements support for syncing to multiple directories for https://github.com/moby/buildkit/pull/2760 as needed when multiple output destinations of type `local` have been specified on the `buildctl` command line.

I'm open to suggestions on how to change/simplify this approach but that's the minimum I came up with.

### DIFF
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -17,7 +18,9 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-type WriteToFunc func(context.Context, string, io.WriteCloser) error
+// WriteToFunc arranges to save data for the file at path
+// to wc
+type WriteToFunc func(ctx context.Context, path string, wc io.WriteCloser) error
 
 type DiskWriterOpt struct {
 	AsyncDataCb   WriteToFunc
@@ -25,13 +28,15 @@ type DiskWriterOpt struct {
 	NotifyCb      func(ChangeKind, string, os.FileInfo, error) error
 	ContentHasher ContentHasher
 	Filter        FilterFunc
+
+	rewriteMetadata func(path string, st *types.Stat) error
 }
 
 type FilterFunc func(string, *types.Stat) bool
 
 type DiskWriter struct {
-	opt  DiskWriterOpt
-	dest string
+	opt   DiskWriterOpt
+	dests []string
 
 	ctx    context.Context
 	cancel func()
@@ -40,11 +45,18 @@ type DiskWriter struct {
 }
 
 func NewDiskWriter(ctx context.Context, dest string, opt DiskWriterOpt) (*DiskWriter, error) {
+	return NewDiskWriterMultiple(ctx, []string{dest}, opt)
+}
+
+func NewDiskWriterMultiple(ctx context.Context, dests []string, opt DiskWriterOpt) (*DiskWriter, error) {
 	if opt.SyncDataCb == nil && opt.AsyncDataCb == nil {
 		return nil, errors.New("no data callback specified")
 	}
 	if opt.SyncDataCb != nil && opt.AsyncDataCb != nil {
 		return nil, errors.New("can't specify both sync and async data callbacks")
+	}
+	if opt.rewriteMetadata == nil {
+		opt.rewriteMetadata = rewriteMetadata
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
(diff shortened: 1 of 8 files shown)


diskwriter.go
```diff
@@ -345,3 +398,83 @@ func nextSuffix() string {
...
+		wcs:    wcs,
+	}
+}
+
+type multiWriter struct {
+	io.Writer
+	wcs []io.WriteCloser
+}
+
+func newMultiErr(errs ...error) error {
```
a-palchikov, line 459: I did not attempt to introduce another dependency on something like [go-multierror](github.com/hashicorp/go-multierror) so this is ad-hoc multiple errors wrapper.

diskwriter_test.go
```diff
@@ -158,17 +158,16 @@ func TestWalkerWriterSimple(t *testing.T) {
 	err = Walk(context.Background(), dest, nil, bufWalk(b))
 	assert.NoError(t, err)
 
-	assert.Equal(t, string(b.Bytes()), `dir bar
+	assert.Equal(t, `dir bar
```
a-palchikov, line 161: Swapped the arguments to match the error message better - expected is the first followed by actual. Otherwise, the error message on failure is confusing.

receive.go
```diff
@@ -26,15 +26,18 @@ type ReceiveOpt struct {
 	Merge         bool
 	Filter        FilterFunc
 	Differ        DiffType
+
+	diskWriterOpt DiskWriterOpt
```
a-palchikov, line 30: added explicit configuration attribute to work around the necessity to run the test as root (see below).

receive_test.go
```diff
@@ -67,8 +67,6 @@ func TestInvalidExcludePatterns(t *testing.T) {
 }
 
 func TestCopyWithSubDir(t *testing.T) {
-	requiresRoot(t)
```
a-palchikov, line 70: Instead of running as root, the test records changes to the file metadata.

receive_test.go
```diff
@@ -178,7 +240,7 @@ func TestCopySimple(t *testing.T) {
 		"ADD zzz/bb dir",
 		"ADD zzz/bb/cc dir",
 		"ADD zzz/bb/cc/dd symlink ../../",
-		"ADD zzz.aa zzdata",
+		"ADD zzz.aa file zzdata",
```
a-palchikov, line 243: This was confusing - was the omission of the file type intended? Let me know if this is the case and how to interpret the test then. I had to add the file type as I updated the changes language parser to panic on unexpected input.

receive_test.go
```diff
@@ -112,10 +174,10 @@ func TestCopySwitchDirToFile(t *testing.T) {
 
 	dest, err := tmpDir(changeStream([]string{
 		"ADD foo dir",
-		"ADD foo/bar dile data2",
+		"ADD foo/bar file data2",
```
a-palchikov, line 177: Was this a typo or intended?

validator_test.go
```diff
@@ -231,6 +231,8 @@ func parseChange(str string) *change {
 		}
 		st.Mode |= uint32(os.ModeSymlink)
 		st.Linkname = f[3]
+	default:
+		panic(fmt.Sprint("unrecognized file type:", typ))
```
a-palchikov, line 235: Added this to find suspected typos in input (see above). If this is intended - please let me know.

receive.go
```diff
@@ -26,15 +26,18 @@ type ReceiveOpt struct {
...
 }
 
 func Receive(ctx context.Context, conn Stream, dest string, opt ReceiveOpt) error {
-	ctx, cancel := context.WithCancel(ctx)
```
a-palchikov, line 32: This looked like a no-op - was it intended?

receive.go
```diff
@@ -139,11 +143,11 @@ func (r *receiver) run(ctx context.Context) error {
 				r.conn.SendMsg(&types.Packet{Type: types.PACKET_ERR, Data: []byte(retErr.Error())})
 			}
 		}()
-		destWalker := emptyWalker
-		if !r.merge {
-			destWalker = getWalkerFn(r.dest)
+		if r.merge {
+			err = createDir(ctx, dw.HandleChange, w.fill)
+		} else {
+			err = doubleWalkDiff(ctx, dw.HandleChange, getWalkerFn(r.dests[0]), w.fill, r.filter, r.differ)
```
tonistiigi, line 149: Not sure how this works if the comparison for the differ is always based on the first path?

tonistiigi (commented):
Do you have a good use-case in mind for this? It seems quite complicated and I don't consider this aspect a blocker for https://github.com/moby/buildkit/pull/2760 . Or I'm also fine if this edge case is handled non-optimally by just transferring things twice.

a-palchikov:
I have to agree to complexity. I will rewrite it as a corner case of copying twice (or however many times) which I was hesitating to implement. Thanks for your feedback!

PR #139, opened 2022-11-03
receiver: re-apply directory timestamps after copying

jedevc:
:hammer_and_wrench: Fixes the bug reported in https://github.com/moby/buildkit/issues/2884#issuecomment-1301028540.

This is needed as creating files in directories modifies their timestamp, so we need to set the timestamp again after creating all the directory contents.

This logic is similar to the logic in [diffapply_unix.go in moby/buildkit](https://github.com/moby/buildkit/blob/master/snapshot/diffapply_unix.go#L142).

### DIFF
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"hash"
 	"io"
+	gofs "io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


tonistiigi (commented):
Could we have a regression test for this?

Are you sure this is the same as issue in the buildkit report? That one seems to be using the copy package.

@smira

jedevc:
Added a regression test.

tonistiigi (approved):

tonistiigi:
Added follow-up in https://github.com/tonistiigi/fsutil/issues/141

PR #140, opened 2022-11-03
copy: restore parent directory timestamp after copying contents

jedevc:
:hammer_and_wrench: Fixes https://github.com/moby/buildkit/issues/2884

We need to restore the previous directory `FileInfo` if it existed prior to attempting to copy content into it.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -294,6 +294,10 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 	if err != nil {
 		return errors.Wrapf(err, "failed to stat %s", src)
 	}
+	targetFi, err := os.Lstat(target)
+	if err != nil && !os.IsNotExist(err) {
+		return errors.Wrapf(err, "failed to stat %s", src)
+	}
 
 	include := true
 	var (
@@ -335,6 +339,7 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 	}
 
 	copyFileInfo := include
+	restoreFileTimestamp := false
 	notify := true
 
 	switch {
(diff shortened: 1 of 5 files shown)


copy/copy.go
```diff
@@ -386,6 +393,10 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 		if err := copyXAttrs(target, src, c.xattrErrorHandler); err != nil {
 			return errors.Wrap(err, "failed to copy xattrs")
 		}
+	} else if restoreFileInfo && targetFi != nil {
+		if err := c.copyFileInfo(targetFi, target); err != nil {
```
tonistiigi, line 397: Could we separate out the timestamp update and only call that. Performance isn't the main issue, but I think it makes the code easier to follow if it is clear in here that we call this just to restore the timestamp.

copy/copy.go
```diff
@@ -386,6 +393,11 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
 		if err := copyXAttrs(target, src, c.xattrErrorHandler); err != nil {
 			return errors.Wrap(err, "failed to copy xattrs")
 		}
+	} else if restoreFileTimestamp && targetFi != nil {
+		t := targetFi.ModTime()
+		if err := Utimes(target, &t); err != nil {
```
tonistiigi, line 398: This doesn't set the atime and mtime separately like the `copyFileInfo`. Not a problem?

jedevc: Yup, probably a problem :laughing:

jedevc: Fixed. Have extracted the timestamp portions of `copyFileInfo` into `copyFileTimestamp`, so we can call that here.

tonistiigi (approved):

Issue #142, opened 2022-11-11
Add Windows CI workflow


gabriel-samfira:
The tests on Windows for a number of packages seem to fail. A lot of them are a simple fix of just doing a ```filepath.FromSlash()``` on expected paths, but some are legitimately failing. Looking at the tested code that fails. Looks like most of it is in regards to following symlinks or walking paths.

As this code is being used in buildkit in https://github.com/moby/buildkit/blob/master/cache/util/fsutil.go#L99, this will have to be fixed before we can safely add Windows support to builkit.

Looking into it now, and will post back an update as soon as I have something.

PR #143, opened 2022-11-17
Fix tests on windows and add workflow

gabriel-samfira:
This change adds a new workflow to test this project on Windows. It also fixes tests and some functionality.

  * Path separator is converted to whatever is specific to the OS we are running on
  * containerd/continuity package updated to latest commit. In newer versions of continuity, ```ErrNotSupported``` is ignored when resolving ```xattrs```, which is unsupported on Windows.
  * UIDs and GIDs are ignored on Windows
  * Add workflows for Windows
  * Disable some tests on Windows

Fixes: #142

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"hash"
 	"os"
-	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/tonistiigi/fsutil/types"
@@ -33,7 +32,7 @@ func getWalkerFn(root string) walkerFn {
 			}
 
 			p := &currentPath{
-				path: filepath.FromSlash(path),
+				path: path,
 				stat: stat,
 			}
 
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"hash"
 	"io"
+	gofs "io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -33,10 +34,11 @@ type DiskWriter struct {
 	opt  DiskWriterOpt
 	dest string
 
-	ctx    context.Context
-	cancel func()
-	eg     *errgroup.Group
-	filter FilterFunc
+	ctx         context.Context
+	cancel      func()
+	eg          *errgroup.Group
+	filter      FilterFunc
+	dirModTimes map[string]int64
 }
 
 func NewDiskWriter(ctx context.Context, dest string, opt DiskWriterOpt) (*DiskWriter, error) {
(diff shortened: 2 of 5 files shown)


gabriel-samfira:
Noticed a couple of bugs. Fixing and committing soon.

gabriel-samfira:
This is ready for review when you get a chance. The bug I noticed was in ```copyFileInfo()```. I had forgotten to use the chowner when defined and kept copying the original file ownership information.

As a side note, I have been investigating why ownership information doesn't carry from layer to layer. Only DACLs seem to carry over. This seems to be a limitations in ```HCS```, not in any library or application code, and happens when layers get activated.

It's worth having this code added, however, as once that behavior is fixed in HCS, we would already have coverage in buildkit/containerd/moby. Also, DACLs seem to be retained, which should allow users to retain whatever access they had to the file regardless of who owns it.

gabriel-samfira:
Hi!

Does anyone have some time to review these changes? :smile:

diff.go
```diff
@@ -32,7 +33,7 @@ func getWalkerFn(root string) walkerFn {
 			}
 
 			p := &currentPath{
-				path: path,
+				path: filepath.FromSlash(path),
```
tonistiigi, line 36: I don't think this is correct if it is compared to the `stat` that is coming from protobuf.

All the paths on the proto wire and public types in this library should be unix style. System variant is only used when opening a file from local system (I think this is all internal behavior). When there is communication to the remote side via send/receive/diskwriter etc., what OS remote side is should be irrelevant.

tonistiigi (commented):
Sorry, I don't have notifications set up for this repo.

Left a couple of quick comments. If the path conversions are needed in existing files we might need to do it in separate commits or prs so they can be evaluated more clearly if the use-case requires normalized path or current system path.

tonistiigi:
Feel free to move the functionally separate parts to smaller PRs for easier review.

receive_test.go
```diff
@@ -98,42 +99,6 @@ func TestCopyWithSubDir(t *testing.T) {
 	assert.Equal(t, "data1", string(dt))
 }
 
-func TestCopyDirectoryTimestamps(t *testing.T) {
```
tonistiigi, line 101: Also looks removed

gabriel-samfira: I can't really explain what malfunction I had that made me decide to remove this. Added it back, as well as the mod times.

diskwriter.go
```diff
@@ -53,32 +51,17 @@ func NewDiskWriter(ctx context.Context, dest string, opt DiskWriterOpt) (*DiskWr
 	eg, ctx := errgroup.WithContext(ctx)
 
 	return &DiskWriter{
-		opt:         opt,
-		dest:        dest,
-		eg:          eg,
-		ctx:         ctx,
-		cancel:      cancel,
-		filter:      opt.Filter,
-		dirModTimes: map[string]int64{},
```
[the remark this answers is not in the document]
gabriel-samfira: They shouldn't be. Added them back. My apologies!

diff.go
gabriel-samfira: Reverted the change. It was handled elsewhere.

gabriel-samfira:
I opened  #145 #146 #147 #148 and #149 to replace this PR. Closing this one. FreeBSD tests seem to fail at setting up vagrant. Probably a github issue.

PR #146, opened 2023-02-09
Implement proper renameFile on Windows

gabriel-samfira:
Symlinks on Windows are implemented using NTFS reparse points. Reparse points are also used when creating volume mounts on Windows. As far as golang is concerned, mountpoints and symlinks look the same. This change makes the distinction between the two and allows safely renaming symlinks, while erring out for anything else that looks like a symlink.

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>

### DIFF
--- a/diskwriter_windows.go
+++ b/diskwriter_windows.go
@@ -5,7 +5,7 @@ package fsutil
 
 import (
 	"fmt"
-	ioFS "io/fs"
+	iofs "io/fs"
 	"os"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


diskwriter_windows.go
```diff
@@ -1,8 +1,15 @@
+//go:build windows
 // +build windows
 
 package fsutil
 
 import (
+	"fmt"
+	ioFS "io/fs"
```
tonistiigi, line 8: what's up with this package alias. `fs` doesn't even seem to be used elsewhere in the file.

gabriel-samfira: remnant from splitting up the previous PR. Changing it now.

gabriel-samfira: The only way we can import ```io/fs``` is if we rename that struct. I think the aliased import is probably the path with the least amount of code changes :smile: . What do you think?

tonistiigi: Ok, but make it all lowercase.

gabriel-samfira: Import was renamed. Please take another look.

tonistiigi (approved):

PR #147, opened 2023-02-09
Fix follow symlinkResolver on Windows

gabriel-samfira:
This change handles Windows absolute paths and properly handles IsNotFound errors.

This change also fixes tests on Windows.

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>

### DIFF
--- a/followlinks.go
+++ b/followlinks.go
@@ -19,7 +19,7 @@ func FollowLinks(root string, paths []string) ([]string, error) {
 	}
 	res := make([]string, 0, len(r.resolved))
 	for r := range r.resolved {
-		res = append(res, r)
+		res = append(res, filepath.ToSlash(r))
 	}
 	sort.Strings(res)
 	return dedupePaths(res), nil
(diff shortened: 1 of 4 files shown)


followlinks.go
```diff
@@ -31,7 +31,13 @@ type symlinkResolver struct {
 }
 
 func (r *symlinkResolver) append(p string) error {
-	p = filepath.Join(".", p)
+	p = filepath.FromSlash(p)
```
tonistiigi, line 34: I think after `FollowLinks` returns, all the return paths should be in unix-style again. Wdyt?

gabriel-samfira: I am trying to think of a case where this code runs on Windows and would end up mutating paths that would eventually end up on a non Windows system. If there is no code path that would make that happen, I am not sure it's worth converting them to Unix style only to convert them back to Windows style later.

Besides sparing us from using ```filepath.FromSlash()``` in the tests, I can't (yet) think of such a case. But I can call ```filepath.ToSlash()``` lower in this function where we add them to ```r.resolved``` if you wish.

tonistiigi: I think even in the current walker code the where this result goes to includepatterns, the `IncludePatterns` value normally would be unix style.

It is also just consistency as it is a public API. A client can call `FollowLinks` manually and then use the result paths as inputs to other functions that afiacs all take unix-style paths.

gabriel-samfira: I think that should do it. Please take another look.

tonistiigi (approved):

PR #153, opened 2023-02-15
Fix leaking file handle

gabriel-samfira:
Remove the break statement that prevented the file from being closed.


gabriel-samfira:
@tonistiigi friendly reminder

gabriel-samfira:
@tonistiigi this fixes a failure on Windows (less forgiving than Linux when it comes to open files) and prevents a file handle leak on Linux.

tonistiigi (approved):

PR #157, opened 2023-04-07
Fix potential nil pointer dereference

gabriel-samfira:
We need to check that the user is not nil before de-referencing it.


gabriel-samfira:
CC: @tonistiigi

tonistiigi (approved):

PR #158, opened 2023-04-07
Use RunWithPrivileges

gabriel-samfira:
The ```EnableProcessPrivileges()``` function will enable a certain set of privileges for the entire process. I am unsure if enabling/disabling privileges is tracked with some sort of reference counting, but if it's not, when we release those privileges, we disable them for the entire process. This means that if some other go routine requires them, we may end up with undesirable results.

The ```RunWithPrivileges()``` function locks the thread in which the function runs, and enables those privileges only in that thread.

### DIFF
--- a/copy/mkdir_windows.go
+++ b/copy/mkdir_windows.go
@@ -54,16 +54,6 @@ func Chown(p string, old *User, fn Chowner) error {
 		userSIDstring = containerAdministratorSidString
 
 	}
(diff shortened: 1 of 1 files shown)


gabriel-samfira:
The failure seems to be a CI issue.

gabriel-samfira:
CC @tonistiigi

tonistiigi (approved):

PR #160, opened 2023-06-16
walk: avoid stat()'ing files unnecessarily

nicks:
Before:

```
go test -bench=. .
goos: linux
goarch: amd64
pkg: github.com/tonistiigi/fsutil
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkWalker/[1]-target-8         	   27994	     41254 ns/op
BenchmarkWalker/[1]-**/target-8      	   29540	     40402 ns/op
BenchmarkWalker/[2]-*/target-8       	    1017	   1104834 ns/op
BenchmarkWalker/[2]-**/target-8      	     920	   1124731 ns/op
BenchmarkWalker/[3]-*/*/target-8     	      34	  40524094 ns/op
BenchmarkWalker/[3]-**/target-8      	      32	  40432908 ns/op
BenchmarkWalker/[4]-*/*/*/target-8   	      31	  44836714 ns/op
BenchmarkWalker/[4]-**/target-8      	      26	  44611379 ns/op
BenchmarkWalker/[5]-*/*/*/*/target-8 	      85	  14179975 ns/op
BenchmarkWalker/[5]-**/target-8      	      78	  13620031 ns/op
BenchmarkWalker/[6]-*/*/*/*/*/target-8         	      44	  23380013 ns/op
BenchmarkWalker/[6]-**/target-8                	      51	  22435264 ns/op
BenchmarkWalker/[6]-**-!*/*/**-8               	    2101	    580936 ns/op
```

After:

```
go test -bench=. .
goos: linux
goarch: amd64
pkg: github.com/tonistiigi/fsutil
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkWalker/[1]-target-8         	   40788	     26219 ns/op
BenchmarkWalker/[1]-**/target-8      	   41642	     26998 ns/op
BenchmarkWalker/[2]-*/target-8       	    1725	    660271 ns/op
BenchmarkWalker/[2]-**/target-8      	    1806	    645525 ns/op
BenchmarkWalker/[3]-*/*/target-8     	      64	  17609101 ns/op
BenchmarkWalker/[3]-**/target-8      	      66	  17563334 ns/op
BenchmarkWalker/[4]-*/*/*/target-8   	      40	  26241578 ns/op
BenchmarkWalker/[4]-**/target-8      	      43	  24575370 ns/op
BenchmarkWalker/[5]-*/*/*/*/target-8 	     121	   9564283 ns/op
BenchmarkWalker/[5]-**/target-8      	     126	   9412483 ns/op
BenchmarkWalker/[6]-*/*/*/*/*/target-8         	      63	  18703020 ns/op
BenchmarkWalker/[6]-**/target-8                	      73	  17287584 ns/op
BenchmarkWalker/[6]-**-!*/*/**-8               	    3792	    273148 ns/op
```


stat.go
```diff
@@ -1,6 +1,7 @@
 package fsutil
 
 import (
+	gofs "io/fs"
```
thaJeztah, line 4: Curious; was there a conflict somewhere that made an alias needed? (Otherwise, I think it'd be good to avoid aliases)

thaJeztah: Actually wondering if the change is needed, as `os.FileInfo` is an alias for `fs.FileInfo` (and `os.FileInfo` has not been deprecated 🤔  https://github.com/golang/go/blob/1a7709d6af76c06d465c5e969b502fc206f8e687/src/os/types.go#L21

```go
// A FileInfo describes a file and is returned by Stat and Lstat.
type FileInfo = fs.FileInfo
```

nicks: ya, there's a conflict with the `fs` type in this package:
https://github.com/tonistiigi/fsutil/blob/9e7a6df4857652bc85225f55e2edb8aece3ecd43/fs.go#L29
https://github.com/tonistiigi/fsutil/blob/9e7a6df4857652bc85225f55e2edb8aece3ecd43/diskwriter.go#L7

but you're right, they're the same, i'll change it back.

walker.go
```diff
@@ -161,6 +162,14 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 			}
 
 			if isDir {
+				fi, err := dirEntry.Info()
+				if err != nil {
+					if isNotExist(err) {
+						return nil
```
tonistiigi, line 168: this should be `filepath.SkipDir`?

tonistiigi: If this is the case, then you can just remove this condition as `SkipDir` is handled in the defer funciton.

nicks: ya, i think that would be slightly better

notice that this is handling the narrow case where Readdir() saw a dir, then the dir was removed before we had a chance to walk it.

walker.go
```diff
@@ -268,6 +277,14 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 
 		dir.calledFn = true
 
+		fi, err := dirEntry.Info()
```
tonistiigi, line 280: Can this be a duplicate call for the same file that already had lstat call in 165?

Looks like the value is not cached and always results in a syscall. https://github.com/golang/go/blob/master/src/os/file_unix.go#L472

tonistiigi:
What about a case when a file changed from a directory into, for example, symlink (or vice versa) between the getdents and lstat. Should we handle that case for safety (check that `IsDir()` value did not change after calling `Info()`) and error otherwise? Ideally we would rerun with the value from `Info()` but that may be too complicated change.

walker.go
nicks: ya, good call. it makes a slight difference.

nicks:
re: "What about a case when a file changed from a directory into, for example, symlink (or vice versa) between the getdents and lstat."

if you felt that case was important, i'd probably advocate for an end-to-end fuzz test for it (i.e., something that randomly changed symlinks to directories and vice versa, and make sure the file send does something reasonable). I would feel silly spending a bunch of time handling race conditions in the Walk() code path, only to determine that the race conditions are worse in the send() code path.

walker.go
tonistiigi: Are you planning to make that change? Also not the skipdir handler in this error return.

walker.go
```diff
@@ -161,6 +162,14 @@ func Walk(ctx context.Context, p string, opt *WalkOpt, fn filepath.WalkFunc) err
 			}
 
 			if isDir {
+				fi, err := dirEntry.Info()
+				if err != nil {
+					if isNotExist(err) {
```
tonistiigi, line 167: This extra `if` is not needed here. Already handled in defer.

nicks: oops, my mistake, i misunderstood your comment above

tonistiigi (approved):

tonistiigi:
Looks like a new push is needed to kick the CI. Github had trouble today but looks like it is back now (at least in other repos).

nicks:
it looks like the ci might be broken? i don't see any runs for two weeks https://github.com/tonistiigi/fsutil/actions

PR #161, opened 2023-06-28
handle mkdir race for diskwriter

coryb:
If using the diskwriter hanlders in parallel we might see a common directory being created in parallel, only one will win.  This allows us to gracefully retry/re-evaluate when we get a syscall.EEXIST error on the mkdir instead of returning an error.

### DIFF
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -162,6 +162,10 @@ func (dw *DiskWriter) HandleChange(kind ChangeKind, p string, fi os.FileInfo, er
 	switch {
 	case fi.IsDir():
 		if err := os.Mkdir(newPath, fi.Mode()); err != nil {
+			if errors.Is(err, syscall.EEXIST) {
+				// we saw a race to create this directory, so try again
+				return dw.HandleChange(kind, p, fi, nil)
+			}
 			return errors.Wrapf(err, "failed to create dir %s", newPath)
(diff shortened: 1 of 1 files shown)


tonistiigi:
> If using the diskwriter hanlders in parallel

How is this possible?

coryb:
Sorry, I meant that the logic of the diskwriter handlers could be run in parallel, not that the handler go references were used in parallel.   We are seeing the `directory exists` error  when running buildkit solves in parallel that have common export paths. For example, if the first solve exports to ./build/foo, and the second solve exports to ./build/bar, then there is a race to create ./build if that directory does not pre-exist both solve requests.

tonistiigi:
> when running buildkit solves in parallel that have common export paths.

Ok, I was thinking that this shouldn't be possible on daemon side and it is not safe. But I guess on the client side this is fine as the user is asking to do the best effort. Theoretically, this could result in a root directory escape though if the target directory is changing while one diskwriter is running.

tonistiigi (approved):

tonistiigi:
Looks like a new push is needed to kick the CI. Github had trouble today but looks like it is back now (at least in other repos).

coryb:
Odd, the workflow was disabled, but it let me re-enable it.  image

PR #167, opened 2023-08-02
fs: split out include/exclude into new FilterFS

jedevc:
This patch creates a new FilterFS to pair with a new FilterOpt, which contains the old WalkOpt parameters.

Essentially, this splits the functionality of the old FS into a new FS and the FilterFS. The new FS implemenation simply performs operations over a specified local filesystem, while the FilterFS wraps an existing FS implementation to apply filtering patterns.

This allows the same logic for filtering to apply to *any* underlying filesystem, for example, the StaticFS and MergeFS implementations in BuildKit (see https://github.com/moby/buildkit/pull/4094).

To do this, we also make some reasonably substantial changes to the developer-facing API. We need to use fs.WalkDirFunc instead of the old filepath.WalkFunc, which is required to preserve the lazy stat() semantics from b9e22fc139c520a1339bd31b9abc7c537d9d1623.

Existing implementations and callers of FS should fairly easily be able to update to support the new API, which just requires updating to support a new "path" parameter, and modify the signature to be fs.WalkDirFunc (which may actually improve performance depending on the exact usage).

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -15,12 +15,22 @@ import (
 )
 
 type FilterOpt struct {
+	// IncludePatterns requires that the path matches at least one of the
+	// specified patterns.
 	IncludePatterns []string
+
+	// ExcludePatterns requires that the path does not match any of the
+	// specified patterns.
 	ExcludePatterns []string
-	// FollowPaths contains symlinks that are resolved into include patterns
-	// before performing the fs walk
+
+	// FollowPaths contains symlinks that are resolved into IncludePatterns
+	// at the time of the call to NewFilterFS.
 	FollowPaths []string
-	Map         MapFunc
+
+	// Map is called for each path that is included in the result.
+	// The function can modify the stat info for each element, while the result
+	// of the function controls both how Walk continues.
+	Map MapFunc
 }
 
 type MapFunc func(string, *types.Stat) MapResult
@@ -54,6 +64,16 @@ type filterFS struct {
 	mapFn MapFunc
 }
 
+// NewFilterFS creates a new FS that filters the given FS using the given
+// FilterOpt.
+
+// The returned FS will not contain any paths that do not match the provided
+// include and exclude patterns, or that are are exlcluded using the mapping
+// function.
+//
+// The FS is assumed to be a snapshot of the filesystem at the time of the
+// call to NewFilterFS. If the underlying filesystem changes, calls to the
+// underlying FS may be inconsistent.
 func NewFilterFS(fs FS, opt *FilterOpt) (FS, error) {
 	if opt == nil {
 		return fs, nil
--- a/filter.go
+++ b/filter.go
@@ -44,37 +44,30 @@ const (
 )
 
 type filterFS struct {
-	fs  FS
-	opt *FilterOpt
+	fs FS
+
+	includeMatcher              *patternmatcher.PatternMatcher
+	excludeMatcher              *patternmatcher.PatternMatcher
+	onlyPrefixIncludes          bool
+	onlyPrefixExcludeExceptions bool
+
+	mapFn MapFunc
 }
 
-func NewFilterFS(fs FS, opt *FilterOpt) FS {
+func NewFilterFS(fs FS, opt *FilterOpt) (FS, error) {
 	if opt == nil {
-		return fs
+		return fs, nil
 	}
-	return &filterFS{fs, opt}
-}
 
-func (fs *filterFS) Open(p string) (io.ReadCloser, error) {
-	return fs.fs.Open(p)
-}
-
-func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc) error {
-	var (
-		includePatterns []string
-		includeMatcher  *patternmatcher.PatternMatcher
-		excludeMatcher  *patternmatcher.PatternMatcher
-		err             error
-	)
-
-	if fs.opt != nil && fs.opt.IncludePatterns != nil {
-		includePatterns = make([]string, len(fs.opt.IncludePatterns))
-		copy(includePatterns, fs.opt.IncludePatterns)
+	var includePatterns []string
+	if opt.IncludePatterns != nil {
+		includePatterns = make([]string, len(opt.IncludePatterns))
+		copy(includePatterns, opt.IncludePatterns)
 	}
-	if fs.opt != nil && fs.opt.FollowPaths != nil {
-		targets, err := FollowLinks(fs.fs, fs.opt.FollowPaths)
+	if opt.FollowPaths != nil {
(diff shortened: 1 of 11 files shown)


filter.go
```diff
@@ -43,31 +43,66 @@ const (
...
+}
+
+func NewFilterFS(fs FS, opt *FilterOpt) FS {
+	if opt == nil {
+		return fs
 	}
-	if !rootFI.IsDir() {
-		return errors.WithStack(&os.PathError{Op: "walk", Path: root, Err: syscall.ENOTDIR})
+
+	if fs, ok := fs.(*filterFS); ok {
```
jedevc, line 56: Note to self: need to add some tests to check that this kind of combining is actually valid in all cases.

jedevc: Still need to add some tests, but, I've convinced myself this should be good.

IncludePatterns and ExcludePatterns both set `skip = true` on a match, so it's a purely additive check, so the order really shouldn't matter. The paths stay consistent between Walks, so again, just doing the naive concatenation should be fine.

tonistiigi: This definitely needs tests or if there isn't a performance issue then it can be removed in the initial version. Otherwise, we have 2 competing interfaces that can easily get out of sync.

filter.go
```diff
@@ -43,31 +43,66 @@ const (
...
 	}
 
+	return &filterFS{fs, opt}
+}
+
+func (fs *filterFS) Open(p string) (io.ReadCloser, error) {
+	return fs.fs.Open(p)
+}
+
+func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc) error {
```
tonistiigi, line 92: This `Walk()` method here seems out of place. The signature for walk should be `Walk(fs, .., cb)`. FS should implement something that make this possible, eg. `ReadDir(name)` and `f.ReadDir()`

filter.go
```diff
@@ -43,31 +43,66 @@ const (
...
+			FollowPaths:     follows,
+			Map:             mapFn,
+		}}
 	}
 
+	return &filterFS{fs, opt}
+}
+
+func (fs *filterFS) Open(p string) (io.ReadCloser, error) {
+	return fs.fs.Open(p)
```
tonistiigi, line 89: This isn't really correct iiuc. Eg. when asking for a file that is filtered out it would still be opened.

filter.go
jedevc: The current semantics are difficult to express if we change to that model.

Imagine the following `ExcludePatterns`: (`/foo`, `!/foo/x`).

Currently, if we `Walk` this filesystem, the callback will only be called for `/foo` if `/foo/x` exists, otherwise it gets skipped.

If we try and express `ReadDir` like this, then `ReadDir("/")` will need to check for the existence of `/foo/x` to know if `/foo` should be returned. This would be quite a major rewrite of the include/exclude patterns logic, not sure if you think this would be worth it?

tonistiigi: Maybe the logic then needs to be that filter walks whole baseFS internally either on constructor/function-call or on the first use.

> The current semantics are difficult to express if we change to that model.

I see, but I don't see the benefit then of defining a `FilterFS` if it doesn't actually satisfy the `FS` interface.

filter.go
jedevc: This actually isn't consistent, so we need to remove this additional check. See the new test, it's not valid to combine inclusions/exclusions as shown.

filter.go
jedevc: https://github.com/tonistiigi/fsutil/blob/36ef4d8c0dbba81b41fcfe7a03ba1f6ba56904f2/fs.go#L17-L20

It does though? `fsutil.FS` only defines `Open` and `Walk` for now.

I think switching to be more in match with `io/fs` is a good idea, but I'd propose that's out of scope for this PR - unless we want to make the argument to do a large PR to completely rework the internals of this package.

filter_test.go
```diff
@@ -708,6 +754,119 @@ func TestWalkerDoublestarInclude(t *testing.T) {
...
+		"ADD foo file",
+		"ADD bar file",
+	}))
+	assert.NoError(t, err)
+	defer os.RemoveAll(d)
+	f, err := NewFS(d)
+	assert.NoError(t, err)
+
+	f, err = NewFilterFS(f, &FilterOpt{
+		ExcludePatterns: []string{"bar"},
```
tonistiigi, line 849: Could we do something more complicated with wildcards and `!`.

followlinks.go
```diff
@@ -99,21 +102,30 @@ func (r *symlinkResolver) readSymlink(p string, allowWildcard bool) ([]string, e
 		return out, nil
 	}
 
-	fi, err := os.Lstat(realPath)
+	entry, err := readFile(r.fs, p)
```
tonistiigi, line 105: readFile returns `DirEntry`?

followlinks.go
```diff
@@ -122,6 +134,80 @@ func (r *symlinkResolver) readSymlink(p string, allowWildcard bool) ([]string, e
...
 
+func readFile(fs FS, root string) (os.DirEntry, error) {
+	var out os.DirEntry
+
+	root = filepath.Clean(root)
+	if root == "/" || root == "." {
+		return nil, nil
+	}
+
+	err := fs.Walk(context.TODO(), root, func(p string, entry os.DirEntry, err error) error {
```
tonistiigi, line 145: is `fs.Walk` called on non-directory here?

jedevc: As discussed, this should be fine, since we're just doing the same behavior from the go standard library.

followlinks.go
jedevc: Renamed to `statFile`

followlinks.go
```diff
@@ -122,6 +134,80 @@ func (r *symlinkResolver) readSymlink(p string, allowWildcard bool) ([]string, e
...
+	if root == "/" || root == "." {
+		root = "."
+		out = make([]gofs.DirEntry, 0)
+	}
+
+	err := fs.Walk(context.TODO(), root, func(p string, entry os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		rel, err := filepath.Rel(root, p)
```
tonistiigi, line 181: why the `Rel` instead of plain equality check?

tonistiigi (commented):
I'm still not quite sure why implement `statFile` with `Walk` instead of defining `Stat()` (and then maybe removing path from `Walk`) but that discussion can be left for some follow-up.

followlinks.go
jedevc: An oversight, fixed.

jedevc:
@tonistiigi is there anything blocking this? Happy to take another pass (or prep a follow-up) if there's anything needed here.

tonistiigi (approved):

PR #168, opened 2023-09-25
copy: fix to respect most significant octet of given permission in copyFileInfo

musaprg:
This PR fixes `(*copier).copyFileInfo()` to respect special bits specified in the copier struct. It causes [the BuildKit issue](https://github.com/moby/buildkit/issues/3920) where `COPY --chmod` doesn't respect the most significant octet of the Unix Permission.

This behavior might be intended because setting setuid/setgid to an executable file was known as an anti-pattern in terms of security. However, this command should not ignore them implicitly if the user understands its drawbacks and desires to set them.

## Additional Information

It seems Go's `os.FileMode` has its internal representation of special bits. We need to convert special bits to internal representation to set them properly. 

> Additionally, when os.Chmod() runs, it [actively discards](https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/os/file_posix.go;l=62-88) anything past the first 9 bits (owner, group, other / read, write, execute) and uses its own internal representation to set the special file mode bits.
https://github.com/coreos/ignition/issues/1301#issue-1099348447

### DIFF
--- a/copy/copy_linux.go
+++ b/copy/copy_linux.go
@@ -30,12 +30,15 @@ func (c *copier) copyFileInfo(fi os.FileInfo, src, name string) error {
 
 	m := fi.Mode()
 	if c.mode != nil {
-		m = os.FileMode(*c.mode & 0777)
-		sb := *c.mode & 0777
-		if sb != 0 {
-			m |= os.FileMode(sb)
-		} else {
-			m |= m & ^os.FileMode(0777)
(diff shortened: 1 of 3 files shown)


musaprg:
It seems CI failed due to another issue on freebsd job, which might not be related to this PR.

I suppose we need to wait for this patch being merged into master. https://github.com/tonistiigi/fsutil/pull/172

tonistiigi (approved):

PR #169, opened 2023-11-06
Add support for always overwriting existing paths

sipsma:
Before this, copy would always refuse to replace an existing path in the dest with something from the source of a different type. E.g. if there was a symlink at the dest, you'd get an error if copying src meant replacing the symlink w/ a directory.

This is the right default behavior but there are situations in which callers may want to opt in to instead always replacing what's in dest with what's in source.

This adds an option to enable that behavior.

---

cc @tonistiigi @jedevc PTAL, if we're good with adding this option I'll follow up with adding this option to LLB's `CopyInfo` too.

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -88,7 +88,7 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


sipsma:
ping @tonistiigi

tonistiigi (approved):

PR #171, opened 2023-11-16
Migrate off of gogo/protobuf

kzys:
gogo/protobuf has been deprecated since 2022. This change uses Google's code generator instead of gogo/protobuf.


clone.go
```diff
@@ -0,0 +1,18 @@
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func cloneProto[T protoreflect.ProtoMessage](from T) (T, error) {
+	var zero T
+	clone := proto.Clone(from)
+	typed, ok := clone.(T)
+	if !ok {
+		return zero, fmt.Errorf("failed to cast %+v to %+v", clone, zero)
```
tonistiigi, line 15: only `pkg/errors` allowed in these repos

clone.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func cloneProto[T protoreflect.ProtoMessage](from T) (T, error) {
+	var zero T
+	clone := proto.Clone(from)
```
tonistiigi, line 12: I'm a bit worried about including this reflect-based clone in this internal logic where no reflect should be needed. Are there other options?

tonistiigi: Eg. maybe some of these private functions should not use proto based types.

clone.go
kzys: I didn't know that. Will fix.

clone.go
kzys: Ended up just doing the below.

```
	clone := &types.Stat{
		Path:     from.Path,
		Mode:     from.Mode,
		Uid:      from.Uid,
		...
	}
```

thaJeztah:
I triggered CI; looks like there's a small linting failure;

```
17.56 receive_test.go:389:6: func `clonePacket` is unused (unused)
17.56 func clonePacket(from *types.Packet) *types.Packet {
17.56      ^
```

kzys:
@thaJeztah Thanks! Fixed.

thaJeztah:
Awesome; I kicked it again, and it's green now 👍 

(I'm not super-familiar with this code, so I'll leave actual review to @tonistiigi, but noticed the PR, and saw CI needed approval 😅)

tonistiigi (approved):
Do we need some benchmarks as well to ensure there isn't a big regression in marshal/unmarshal? I think in BuildKit I would like some tests that verify backwards compatibility as well (by including binary proto results from previous version).

PR #183, opened 2024-02-19
Fix `FilterFS` when skipping lazy parent directories

jedevc:
Original issue found by @vito in https://github.com/dagger/dagger/issues/6687.

Split into two commits, both with slightly different issues:
- `filter: allow SkipDir return with lazy parents`. Ensures that walker functions that return `filepath.SkipDir` are always respected, even when evaluated as a lazy parent function (see commit for more details).
- `filter: ensure MapResult is followed for parent directories`. Ensures that the `FilterFS` Map option for `MapResultSkipDir` is always respected, even when evaluated as a lazy parent function (see commit for more details).

These issues seem to have been present for a while (including prior to #167, which introduced `FilterFS`, where the issue was still present in `fsutil.Walk`).
- The first issue seems to have appeared in https://github.com/tonistiigi/fsutil/pull/108, when lazy parents were introduced.
- The second issues seems to have appeared in https://github.com/tonistiigi/fsutil/pull/130, when map options were introduced.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -176,6 +176,7 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 		includeMatchInfo patternmatcher.MatchInfo
 		excludeMatchInfo patternmatcher.MatchInfo
 		calledFn         bool
+		skipFn           bool
 	}
 
 	// used only for include/exclude handling
@@ -333,6 +334,9 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 				}
 			}
 			for i, parentDir := range parentDirs {
+				if parentDir.skipFn {
+					return filepath.SkipDir
+				}
 				if parentDir.calledFn {
 					continue
 				}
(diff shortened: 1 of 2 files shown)


crazy-max (approved):
LGTM

filter_test.go
```diff
@@ -143,6 +143,30 @@ file bar/foo
...
+
+	err = Walk(context.Background(), d, &FilterOpt{
+		IncludePatterns: []string{"**/*.txt"},
+	}, func(path string, info gofs.FileInfo, err error) error {
+		found = append(found, path)
+		return filepath.SkipDir
+	})
+	assert.NoError(t, err)
+
+	assert.Equal(t, []string{"foo"}, found)
```
tonistiigi, line 167: Why would this return anything at all. "foo" does not match the `IncludePatterns`.

aaronlehmann: `Walk` finds a file that matches the include patterns, then starts calling `Walk` on parent directories, and the first `Walk` callback func returns `SkipDir`...

filter.go
```diff
@@ -356,8 +356,11 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 				}
 				if fs.mapFn != nil {
 					result := fs.mapFn(parentStat.Path, parentStat)
-					if result == MapResultSkipDir || result == MapResultExclude {
+					if result == MapResultExclude {
 						continue
+					} else if result == MapResultSkipDir {
```
aaronlehmann, line 361: nit: don't need `else if` after a `continue`.

filter_test.go
jedevc: Exactly that, yup.

tonistiigi: I don't understand how this makes sense. `IncludePatterns` is an option to `Walk`. Callback should not be called on paths that don't match `IncludePatterns`.

tonistiigi: If the callback calls are considered some "internal implementation specific behaviour" (what is should not I think) then I don't see point in giving the impression that the caller can control the behavior with `SkipDir` (as ultimately some internal logic is not controllable by parameters when determining if callbacks happen).

tonistiigi (approved):

filter_test.go
jedevc: If it's not included there then it isn't hooked up to Send/Recv - which means that we don't send the file stats for parent directories.

e.g. `foo` as this parent should have it's permissions/etc transferred through these methods. Potentially Send shouldn't use Walk at all, but that seems out of scope for this IMO, as it's a larger refactor and architectural change (not really convinced by it either).

PR #187, opened 2024-02-23
send: ensure file path to unix form when sending files

crazy-max:
fixes https://github.com/docker/buildx/issues/2207
closes #186
related to https://github.com/docker/buildx/issues/2207#issuecomment-1957316154

alternative to #186 to make sure unix-style path is sent through the wire so we don't need extra conversion on the receiver side.

cc @jedevc @gabriel-samfira

### DIFF
--- a/send.go
+++ b/send.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


tonistiigi (approved):

tonistiigi:
Need to test in a follow-up if this breaks something for wcow.

jedevc:
Already merged, but LGTM 🎉

PR #191, opened 2024-03-01
test with data races detection

crazy-max:
follow-up https://github.com/tonistiigi/fsutil/pull/164


tonistiigi (commented):
Maybe we should run it on a separate workflow when race is enabled? Enabling race also changes the timing and our users don't run race enabled.

PR #195, opened 2024-04-03
receive: ensure callback errors are propagated

jedevc:
Previously, it was possible for errors returned by an asyncronous callback function to not be propagated, and to instead return "context canceled" from `Receive`.

This could happen based on this specific sequence of events:
- An asyncronous callback is called from `HandleChange`, and returns an error, completing the async errorgroup context with an error state.
- Another call to `HandleChange` completes, and uses the context syncronously, returning `context.Canceled`.
- This in turn propagates to `doubleWalkDiff`, which returns the canceled error, instead of calling `Wait` which would contain the actual error group error.

This behavior is racy, based on exactly how and when the asyncronous callback is called.

The fix to this is relatively straightforward: we just need to make sure that syncronous and asyncronous function calls are kept separate, and don't reuse the same context. This means that a failure in an async callback, doesn't cause later sync ones to fail.

### DIFF
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -37,6 +37,7 @@ type DiskWriter struct {
 	ctx         context.Context
 	cancel      func()
 	eg          *errgroup.Group
+	egCtx       context.Context
 	filter      FilterFunc
 	dirModTimes map[string]int64
 }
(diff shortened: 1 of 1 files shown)


jedevc:
If you're interested in reproing, you can get this to happen by inserting a fake error return into `receiver.asyncDataFunc`, and then running `go test -v -count=100 -run TestCopySimple ./...` - some of the failures will include this error, while others don't propagate it correctly.

tonistiigi (approved):

PR #196, opened 2024-04-03
Fix file transfers from windows to linux

jedevc:
This PR should fix file transfers from Windows to Linux (as is the case with docker-on-windows), and fixes https://github.com/moby/buildkit/pull/4806 https://github.com/moby/buildkit/issues/4741 (cc @profnandaa)

Also fixes https://github.com/dagger/dagger/issues/6923.

There are two parts, split into separate commits. This also could probably do with some tests :smile: 

## Reading unix paths as windows ones

Essentially, this change is wrong (cc @crazy-max): https://github.com/tonistiigi/fsutil/pull/173/files#diff-7dd945edf522324729284de47c3bdb062155cc945eaec235fc564631ae4e91b5

The validator consumes data *directly* off of the wire, which is guaranteed to be in unix-path form (see https://github.com/moby/buildkit/pull/4806#pullrequestreview-1965145911). Therefore, we should avoid using the platform-specific `filepath` package, and instead use the unix-specific `path` package.

## Keeping file map keys consistent

The change here https://github.com/tonistiigi/fsutil/pull/173/files#diff-7e712d320828705efdc283be20c75b6afc94b7327362cdd0e4e96fc6c72e645cR114 causes the `receiver.files` map to be indexed in `asyncDataFunc` by a windows path, but we explicitly store the data using a unix path.

The solution here is to make sure we propagate the *correct* path downwards, always using unix paths.

### DIFF
--- a/diff_containerd.go
+++ b/diff_containerd.go
@@ -111,7 +111,7 @@ func doubleWalkDiff(ctx context.Context, changeFn ChangeFunc, a, b walkerFn, fil
 				if filter != nil {
 					filter(f2.path, &statCopy)
 				}
-				f2copy = &currentPath{path: filepath.FromSlash(f2.path), stat: &statCopy}
+				f2copy = &currentPath{path: f2.path, stat: &statCopy}
 			}
 			k, p := pathChange(f1, f2copy)
 			switch k {
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -183,7 +183,7 @@ func (dw *DiskWriter) HandleChange(kind ChangeKind, p string, fi os.FileInfo, er
 		}
 	default:
 		isRegularFile = true
-		file, err := os.OpenFile(newPath, os.O_CREATE|os.O_WRONLY, fi.Mode()) //todo: windows
+		file, err := os.OpenFile(newPath, os.O_CREATE|os.O_WRONLY, fi.Mode())
 		if err != nil {
 			return errors.Wrapf(err, "failed to create %s", newPath)
 		}
@@ -313,7 +313,7 @@ type lazyFileWriter struct {
 
 func (lfw *lazyFileWriter) Write(dt []byte) (int, error) {
 	if lfw.f == nil {
-		file, err := os.OpenFile(lfw.dest, os.O_WRONLY, 0) //todo: windows
+		file, err := os.OpenFile(lfw.dest, os.O_WRONLY, 0)
 		if os.IsPermission(err) {
 			// retry after chmod
 			fi, er := os.Stat(lfw.dest)
--- a/diskwriter.go
+++ b/diskwriter.go
@@ -98,7 +98,7 @@ func (dw *DiskWriter) HandleChange(kind ChangeKind, p string, fi os.FileInfo, er
 		}
 	}()
 
-	destPath := filepath.Join(dw.dest, filepath.FromSlash(p))
+	destPath := filepath.Join(dw.dest, p)
 
 	if kind == ChangeKindDelete {
 		if dw.filter != nil {
--- a/receive.go
+++ b/receive.go
@@ -1,3 +1,33 @@
+// send.go and receive.go describe the fsutil file-transfer protocol, which
+// allows transferring file trees across a network connection.
+//
+// The protocol operates as follows:
+// - The client (the receiver) connects to the server (the sender).
+// - The sender walks the target tree lexicographically and sends a series of
+//   STAT packets that describe each file (an empty stat indicates EOF).
+// - The receiver sends a REQ packet for each file it requires the contents for,
+//   using the ID for the file (determined as its index in the STAT sequence).
+// - The sender sends a DATA packet with byte arrays for the contents of the
+//   file, associated with an ID (an empty array indicates EOF).
+// - Once the receiver has received all files it wants, it sends a FIN packet,
+//   and the file transfer is complete.
+// If an error is encountered on either side, an ERR packet is sent containing
+// a human-readable error.
+//
+// All paths transferred over the protocol are normalized to unix-style paths,
+// regardless of which platforms are present on either side. These path
+// conversions are performed right before sending a STAT packet (for the
+// sender) or right after receiving the corresponding STAT packet (for the
+// receiver); this abstraction doesn't leak into the rest of fsutil, which
+// operates on native platform-specific paths.
+//
+// Note that in the case of cross-platform file transfers, the transfer is
+// best-effort. Some filenames that are valid on a unix sender would not be
+// valid on a windows receiver, so these paths are rejected as they are
+// received. Additionally, file metadata, like user/group owners and xattrs do
+// not have an exact correspondence on windows, and so would be discarded by
+// a windows receiver.
+
 package fsutil
 
 import (
--- a/receive.go
+++ b/receive.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
+	"syscall"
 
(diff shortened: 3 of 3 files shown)


profnandaa:
Thanks for this. I'll test it out on Windows and get back.

profnandaa:
Vendored and tested against repro for https://github.com/moby/buildkit/issues/4741 -- fixed 👍 

```
buildctl build `
>>  --output type=image,name=docker.nandaa.dev/test,push=false `
>>  --progress plain "-frontend=dockerfile.v0" `
>>  --local context=C:/dev/play/dockerfiles/repro-4741/test1 `
>>  --local dockerfile=C:\dev\play\dockerfiles\repro-4741 --no-cache
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 128B done
#1 DONE 0.0s

#2 [internal] load metadata for mcr.microsoft.com/windows/nanoserver:ltsc2022
#2 DONE 1.5s

#3 [internal] load .dockerignore
#3 transferring context: 2B done
#3 DONE 0.0s

#4 [internal] load build context
#4 transferring context: 61B done
#4 DONE 0.0s

#5 [1/2] FROM mcr.microsoft.com/windows/nanoserver:ltsc2022@sha256:6e6053f0358f9522d2d14693f9bc152f47fe04c82c53dc8c6d127a5a823c8720
#5 resolve mcr.microsoft.com/windows/nanoserver:ltsc2022@sha256:6e6053f0358f9522d2d14693f9bc152f47fe04c82c53dc8c6d127a5a823c8720 0.1s done
#5 CACHED

#6 [2/2] COPY . C:/
#6 DONE 0.4s

#7 exporting to image
#7 exporting layers
#7 exporting layers 0.8s done
#7 exporting manifest sha256:4307cc1366f3d306f410a381cdbecb47d4dc19f04616f1a2ecfeecb0593cc1ca 0.0s done
#7 exporting config sha256:a50a7ba21691b82ea2dfef10f82ae4a4012c0f6924aa7e45750e06d98c8894a2 0.0s done
#7 naming to docker.nandaa.dev/test done
#7 DONE 0.9s
```

profnandaa (approved):
LGTM.

crazy-max (approved):
> Essentially, this change is wrong (cc @crazy-max): https://github.com/tonistiigi/fsutil/pull/173/files#diff-7dd945edf522324729284de47c3bdb062155cc945eaec235fc564631ae4e91b5
> 
> The validator consumes data _directly_ off of the wire, which is guaranteed to be in unix-path form (see [moby/buildkit#4806 (review)](https://github.com/moby/buildkit/pull/4806#pullrequestreview-1965145911)). Therefore, we should avoid using the platform-specific `filepath` package, and instead use the unix-specific `path` package.

I picked these changes from https://github.com/tonistiigi/fsutil/pull/148/files#diff-7dd945edf522324729284de47c3bdb062155cc945eaec235fc564631ae4e91b5 but overlooked it was done on receiver side.

LGTM thanks

validator.go
```diff
@@ -27,18 +27,18 @@ func (v *Validator) HandleChange(kind ChangeKind, p string, fi os.FileInfo, err
 	if v.parentDirs == nil {
 		v.parentDirs = make([]parent, 1, 10)
 	}
-	if p != filepath.Clean(p) {
+	if p != path.Clean(p) {
```
tonistiigi, line 30: I think we somehow need to make sure that there are no `\` in this path (at least for wcow). Otherwise we do the validation only based on `/` but then in diskwriter these are converted to local paths in `FromSlash()` and and the previous `\` become active.

Eg. lets say remote party sends:

```
dir foo
symlink foo/sym -> /
dir file foo\sym
file foo\sym/bar
```

I think that would pass validator atm. but would result in breakout in diskwriter.

tonistiigi (approved):

tonistiigi (changes requested):
^

validator.go
tonistiigi: `sym\bar` is a valid filename in linux but I guess can't be in windows. So if validator/diskwriter run on windows then the paths should not contain `\` and it should cause a validator error if there is such path. In Linux, if there is `\` then it should be just written out verbose without making a directory separator.

If there is a backwards compatibility issue then it would also be possible that in wcow we convert all `\` to `/` before we run the validator.

jedevc: Added a check for this case now - the check is actually quite simple - we just convert from and to the slash format, and if there's a difference, the path can't be represented on the target.

This detects the path "lossiness".

---

Note: unlike I mentioned above, I don't think the reverse is an issue, since `/` isn't a valid filename character in windows? So we would never be in a scenario of trying to send a file `foo/bar\baz` from windows to linux (so the reverse issue doesn't apply).

tonistiigi (commented):
Still somewhat confusing to me how this works.

Afaics there is no guarantee that `fs.Walk` returns unix-only paths. The conversion seems to be in `send.go`. On the receiver side `getWalkerFn` uses `fs.Walk` that is then then sent to `ComparePath` through differ that now with this implementation only understands `/` separators.

The same seems to be true for the remote paths via `newDynamicWalker` as well. The validator does do `FromSlash` before validation etc conversion but once it is done, it seems that rest of the functions still use the original form and there is no guarantee what format the path is. This includes the hardlink validation as well as the differ where the format of the path seems to be undefined.

~In diskwriter there is `FromSlash` conversion. It is before the `Filter` gets called. Is this intentional? If yes, it should be documented as such but I think unix-style makes more sense.~

tonistiigi:
I think I'm starting to understand how the differ part works. There is extra `FromSlash` in https://github.com/tonistiigi/fsutil/blob/53090feb8c58f2114893bd0ff1c9c6049ee1f5cf/diff_containerd.go#L114 . But it looks really weird that `doubleWalkDiff` takes two walker functions, first one is expected to be unix style (although I don't think there is validation, per earlier comment) and second one is expected to be `filepath` style and gets converted internally. I think it would make more sense if the conversion for the second one was already part of walker.

For the first walker with files from the remote side, I think the `ToSlash(FromSlash())` is ok, but I think it should be before the validator and the result should rewrite the path for the next functions so all the next ones can be sure that the input is unix-style only.

jedevc:
Okay, so I've changed this pretty substantially - now, we perform translations *immediately* off the wire, and then the rest of the validations/transformations apply in platform-specific style. This keeps validation the same, and means that both funcs in `doubleWalkDiff` now traverse platform paths.

tonistiigi:
@jedevc Can you leave a comment on top of some file explaining where are all the conversions from filepath to unix and unix to filepath, and that the stat receiver guarantees conversion to unix if it is possible.

jedevc:
@tonistiigi have added a high-level description of the protocol (which I've also realized would be useful), and in that described where the path conversions are performed, right before/after sending/receiving. Additionally, I've explained some of the caveats of cross-platform file transfers.

tonistiigi (approved):
I'm not sure how I feel about `Stat` proto type being allowed to be both unix or windows style. I would have preferred if there was a guarantee that this type is only uni unix-style.

But functionally this looks correct, so LGTM.

Things that should be checked in follow-up:
- Filter now doesn't take unix-style path. I think this may be problematic.
- Hardlink detection uses `Linkname == filepath(path)` condition that is likely breaking as it can't be predicted from the client side.

jedevc:
Also for filter ... something seems weird. We pass the *same* `Filter` opt down into into `DiskWriter` and into `doubleWalkDiff`, so the function gets called twice. It definitely doesn't seem like `DiskWriter` needs to be aware of unix-style paths.

I can't quite understand the logic that requires filter to be everywhere like that.

PR #198, opened 2024-04-19
fix hardlink filter regression

tonistiigi:
fix https://github.com/moby/buildkit/issues/4831

With the refactor to FilterFS the hardlink handling was changed so that the hardlink detection is in a separate FS instance and then FilterFS is layered on top of it. This means that the file that was the source for the hardlink could be filtered out, but the Stat pointing to that link is still sent as is.

New function validates if source files are not present in FS anymore and correct the linking. It could be better if all the FS implementation did this automatically, but there is quite a lot of layering going on atm. with multiple layers of FilterFS that would all need to keep own hardlink memory, so atm. the new function is only called before send.

### DIFF
--- a/hardlinks.go
+++ b/hardlinks.go
@@ -1,6 +1,9 @@
 package fsutil
 
 import (
+	"context"
+	"io"
+	gofs "io/fs"
 	"os"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


jedevc (approved):
LGTM, it looks like there are only two `FS` implementations that *remove* files from a base `FS` - `FilterFS` and `SubDirFS`. I think this is probably the easiest fix right now though.

PR #201, opened 2024-06-08
Avoid using syscall for errors on unix

catap:
syscall is deprecated and non updated library whcih should be replaced by new one with name x/sys.

Anyway, it may work on some platform and some errors migth be missed on some. A good example is OpenBSD which has lack of syscall.EBADMSG on amd64 and 386 but has on arm64.

Here I moved all errors to two files: one for unix which uses sys.unix, and one for everything else which uses syscall as fallback.

### DIFF
--- a/errors_nounix.go
+++ b/errors_nounix.go
@@ -1,0 +1,15 @@
+//go:build !unix
+// +build !unix
+
+package fsutil
+
+import "syscall"
+
+const (
+	EBADMSG = syscall.EBADMSG
+	EEXIST  = syscall.EEXIST
+	EINVAL  = syscall.EINVAL
+	EISDIR  = syscall.EISDIR
+	ENOENT  = syscall.ENOENT
+	ENOTDIR = syscall.ENOTDIR
+)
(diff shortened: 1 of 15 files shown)


errors_nounix.go
```diff
@@ -0,0 +1,15 @@
+//go:build !unix
+// +build !unix
+
+package fsutil
```
thaJeztah, line 4: Wondering if these should live in an internal package, as (I think) we don't want consumers to be using these to match against (they're only do prevent having to split out the implementation further, correct?)

tonistiigi: What's the point of returning typed errors if we don't want consumers to match against them?

I don't really see an issue with old code as well, but if we define these separately, then we should export them as well to let consumers no what they can match against.

crazy-max (commented):
@catap There are some conflicts, can you rebase?

catap:
@crazy-max here it is. I haven't tried to build it on go-1.22 but conflict was trivial and grep by patterns `syscall.E` and `'unix.E` doesn't find any new places. So, it should be as fine as it was few months ago

crazy-max (approved):

PR #203, opened 2024-06-26
filter: fix a bug with parent dirs

nicks:
if the Map() function excludes a directory,
but includes a file inside that directory,
then the walker must backtrack and include
the directory.

otherwise, buildkit gets confused and
sends "changes out of order" errors.

part of fixing
https://github.com/tilt-dev/tilt/issues/6393

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -197,7 +197,7 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 			isDir = dirEntry.IsDir()
 		}
 
-		if fs.includeMatcher != nil || fs.excludeMatcher != nil {
+		if fs.includeMatcher != nil || fs.excludeMatcher != nil || fs.mapFn != nil {
 			for len(parentDirs) != 0 {
 				lastParentDir := parentDirs[len(parentDirs)-1].pathWithSep
 				if strings.HasPrefix(path, lastParentDir) {
@@ -298,7 +298,7 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 			return walkErr
 		}
 
-		if fs.includeMatcher != nil || fs.excludeMatcher != nil {
+		if fs.includeMatcher != nil || fs.excludeMatcher != nil || fs.mapFn != nil {
 			defer func() {
 				if isDir {
 					parentDirs = append(parentDirs, dir)
@@ -309,8 +309,6 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 		if skip {
 			return nil
 		}
-
-		dir.calledFn = true
 
 		fi, err := dirEntry.Info()
 		if err != nil {
@@ -333,6 +331,7 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 					return nil
 				}
 			}
+
 			for i, parentDir := range parentDirs {
 				if parentDir.skipFn {
 					return filepath.SkipDir
@@ -354,15 +353,6 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 					return ctx.Err()
 				default:
 				}
-				if fs.mapFn != nil {
-					result := fs.mapFn(parentStat.Path, parentStat)
-					if result == MapResultExclude {
-						continue
-					} else if result == MapResultSkipDir {
-						parentDirs[i].skipFn = true
-						return filepath.SkipDir
-					}
-				}
 
 				parentDirs[i].calledFn = true
 				if err := fn(parentStat.Path, &DirEntryInfo{Stat: parentStat}, nil); err == filepath.SkipDir {
@@ -372,6 +362,8 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 					return err
 				}
 			}
+
+			dir.calledFn = true
 			if err := fn(stat.Path, &DirEntryInfo{Stat: stat}, nil); err != nil {
 				return err
 			}
--- a/filter_test.go
+++ b/filter_test.go
@@ -342,6 +342,7 @@ func TestWalkerMapSkipDirWithPattern(t *testing.T) {
 	assert.NoError(t, err)
 	defer os.RemoveAll(d)
 
+	// The include pattern takes precence over the map function.
 	b := &bytes.Buffer{}
 	err = Walk(context.Background(), d, &FilterOpt{
 		IncludePatterns: []string{"**/*.txt"},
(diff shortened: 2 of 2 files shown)


tonistiigi:
This seems to revert changes from https://github.com/tonistiigi/fsutil/pull/183 (https://github.com/tonistiigi/fsutil/pull/183/commits/d9cfebb97482769495b38d8e14920a4b79b17c95) . From an initial look, the argumentation in #183 makes more sense to me. If `Map` skips the parent dir then it should skip the children of that dir as well, while in the new test, one could make an argument that this is an invalid `Map` function and the function itself should also allow parents if it wants to allow a specific subpath.

@jedevc @aaronlehmann

nicks:
re: "the function itself should also allow parents if it wants to allow a specific subpath"

how would this work? The subpaths are requested from the buildkit server. The Map() function runs on the client, and doesn't know if the subpaths will be requested or not.

Our current workaround is to ALWAYS include the directory if the Map() doesn't know if the subpaths will be requested. The bug downstream (https://github.com/tilt-dev/tilt/issues/6393) is that this means there are sometimes empty folders lying around, which creates problems in python.

tonistiigi:
Maybe that case should use `IncludePatterns` then instead of `Map`. As `Map` is for filtering out files and directories(with their subfiles). Yes, in `Map` if you check `a || a/b || a/b/c` you may end up with a and a/b without any `a/b/c` being present. If you set includepatterns to `a/b/c` then you will not.

nicks:
honestly, poking around at this, I feel like the FilterOpt API doesn't make much sense anymore. it's vestigial from when Map() was a property of DirSource. 

Now that the API supports composable FS implementations, maybe we should just implement our own FS with the Map() semantics we need.

I don't really understand why you like the current Map() semantics -- according to what you're saying, the only way to implement a valid Map() is to have Map() crawl the subdirectories itself, which means it's duplicating all the logic in Walk(). 

is there an example of a project that uses it how you think it should work that you could point me to?

nicks:
ah, poking around at the downstream Dagger issue, i have an idea!

if a Map() returns SkipDir, that should be a "big hammer" that excludes everything underneath. If a subpath of the directory is included by a pattern, the subpath should be ignored.

if a Map() returns Exclude, then that should be softer. If a subpath of the directory is included by a pattern, then the subpath should be included, and the parent backtracker should pick it up again.

i pushed up some tests to demo the new behavior, what do you think?

jedevc:
> honestly, poking around at this, I feel like the FilterOpt API doesn't make much sense anymore. it's vestigial from when Map() was a property of DirSource.

Does `Exclude` even make sense to apply to non-directories? As you mention, the current behavior is bad, and it's possible to end up with an invalid list of files (where you have `a/x.txt`, but not its parent `a/`). Is there any case where you wouldn't just use `SkipDir` instead *today*?

I guess, the idea of this PR is to try and have the map's `Exclude` return value act more inline with the `ExcludePatterns`? If so, I think that seems reasonable.

The change doesn't change the behavior of any existing valid map functions, but simply makes some previously invalid map functions valid (e.g. in the above, we'd get the parent `a/`).

filter.go
```diff
@@ -372,6 +372,8 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 					return err
 				}
 			}
+
+			dir.calledFn = true
```
jedevc, line 376: :eyes: this feels unrelated, were we not setting this before? Was there a case where we were calling the walk target function multiple times? Kinda confused as to why this shows up just here (definitely worth fixing).

nicks: ya, i moved this line.

Before this pr, the logic went:
- check if the directory is skipped by includePatterns or excludePatterns
- set calledFn to true
- check if the directory is skipped by mapFn
- call fn

now, i only set calledFn to true directly before calling fn

nicks:
@jedevc i added some comments on the MapResult options to make the behavior a bit clearer. what do you think?

jedevc (approved):

filter.go
```diff
@@ -197,7 +206,7 @@ func (fs *filterFS) Walk(ctx context.Context, target string, fn gofs.WalkDirFunc
 			isDir = dirEntry.IsDir()
 		}
 
-		if fs.includeMatcher != nil || fs.excludeMatcher != nil {
+		if fs.includeMatcher != nil || fs.excludeMatcher != nil || fs.mapFn != nil {
```
tonistiigi, line 209: One thing I'm not a big fan of is that everything that gets a map goes to this more complicated code-path to support the backtracking, while most of the time the map is really used to reset uid/gid or skip paths recursively.

Would be really ugly, but smth. like `MapAllowBacktrack bool` would solve that.

I don't know how big issue this is in practice though. It might be that most of our requests set includematches/excludematcher anyway.

tonistiigi (commented):
Does this mean that https://github.com/moby/buildkit/blob/v0.15.0-rc1/exporter/local/fs.go#L116 is wrong (or at least needs to be updated after this PR)?

filter.go
nicks: is your concern more about performance or debuggability?

i'd be in favor of adding a much simpler FS implementation purely focused on the reset uid/gid case.

nicks:
re: https://github.com/moby/buildkit/blob/v0.15.0-rc1/exporter/local/fs.go#L116 -- i'm not really sure what the intended behavior of that code is, but i think you would want the backtracking there, no?

filter.go
tonistiigi: Performance a bit maybe, but mostly keeping the default code path simpler and preventing potential stability issues.

tonistiigi:
> , but i think you would want the backtracking there, no?

I don't think so. If a directory causes a remap error then we shouldn't just add it (with wrong permissions) if it has some child under it.

jedevc:
> re: [moby/buildkit@v0.15.0-rc1/exporter/local/fs.go#L116](https://github.com/moby/buildkit/blob/v0.15.0-rc1/exporter/local/fs.go?rgh-link-date=2024-07-09T22%3A09%3A53Z#L116)

The current behavior also seems quite odd to me - if a directory is excluded here, then the files will still be added underneath under the current behavior: which would cause an error on transmission.

Currently directories in this case aren't excluded. I think the "right" thing to do in this case would be to use `MapResultSkipDir` here for directories, I'm not *quite* sure why it isn't already.

nicks:
in theory, i feel like that code should be using SkipDir -- it'd be faster to skip the whole directory, rather than continuing to traverse it

in practice...i couldn't even figure out how to get that codepath to trigger? i traced back to this 5 year old comment about it being an incomplete feature :eyes: https://github.com/moby/buildkit/blob/e83d79a51fb49aeb921d8a2348ae14a58701c98c/cmd/buildkitd/config/config.go#L101, but i'm not sure if that comment is still true

tonistiigi:
> in practice...i couldn't even figure out how to get that codepath to trigger?

This is enabled by the `--userns-remap` feature of dockerd.

nicks:
sorry, have been a bit busy the last couple weeks and finally got back to this.

i think the new behavior i'm introducing in this PR is the correct behavior for tarball exports. I opened a PR in buildkit that demonstrates the issue:
https://github.com/moby/buildkit/pull/5411

PR #206, opened 2024-09-14
deps: remove deprecated gogo proto

jsternberg:
This removes the deprecated gogo proto in favor of the standard implementation.


jsternberg:
Still need to update the bytes for the windows side of the tests but I need to do that on a windows machine. Is there a way we can improve the tests so it isn't reliant on the exact byte pattern for marshal?

jsternberg:
Looks like the hashes didn't change when testing with the CI so going to undo that part of the change. This is ready for review now.

docker-bake.hcl
```diff
@@ -39,6 +39,14 @@ target "lint" {
   }
 }
 
+target "generate" {
```
crazy-max, line 42: we should have another target validating generation similar to buildkit one:
* https://github.com/moby/buildkit/blob/e15601a00fbef2805db1ed87be7bb88628ae926b/hack/dockerfiles/generated-files.Dockerfile#L42-L56
* https://github.com/moby/buildkit/blob/e15601a00fbef2805db1ed87be7bb88628ae926b/docker-bake.hcl#L172-L177

and add it to the validate job matrix: https://github.com/tonistiigi/fsutil/blob/43b9329361d9991d3b945ca13a6ace5f793b134a/.github/workflows/ci.yml#L23-L27

jsternberg: Added this. I also redid part of the generate script because running protoc is very frustrating. We want the import path to be the fully qualified path so that buildkit can import `wire.proto` using the fully qualified path, but protoc was just not happy with the paths being different between projects so I just decided "we're doing this in a dockerfile anyway so I'll just write the protoc command there rather than go generate."

docker-bake.hcl
```diff
@@ -39,6 +39,20 @@ target "lint" {
   }
 }
 
+target "validate-generate" {
```
tonistiigi, line 42: "validate-generated-files"

tonistiigi:
PTAL @kzys

hack/update-generated-files
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail -x
+
+: ${CONTINUOUS_INTEGRATION=}
+
+progressFlag=""
+if [ "$CONTINUOUS_INTEGRATION" == "true" ]; then progressFlag="--progress=plain"; fi
```
crazy-max, line 8: This is not needed as there is no tty allocated on GHA. I think you can just remove this script

hack/validate-generated-files
crazy-max, line 1: Not sure why we need this script, could we not just have it inlined in the dockerfile like https://github.com/moby/buildkit/blob/23a2b6fab3f654e67e4407d5f9ef496ced12f029/hack/dockerfiles/generated-files.Dockerfile#L42-L56?

jsternberg: I just copied what I saw the existing update and validate scripts were doing. We can always do it a different way.

jsternberg: I've updated all of the hack scripts and the dockerfiles to use this inline technique rather than put too much logic in the hack scripts. The hack scripts now just invoke the appropriate bake target and the validate is now inlined in the dockerfile for gomod and generated-files.

go.mod
```diff
@@ -5,21 +5,22 @@ go 1.20
 require (
 	github.com/Microsoft/go-winio v0.5.2
 	github.com/containerd/continuity v0.4.1
-	github.com/gogo/protobuf v1.3.2
 	github.com/moby/patternmatcher v0.5.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/pkg/errors v0.9.1
+	github.com/planetscale/vtprotobuf v0.6.0
```
crazy-max, line 11: do we want to separate vtproto as well here related to https://github.com/moby/buildkit/pull/5342#issuecomment-2371804240?

jsternberg:
Removed the vtproto part of this and I'll create a separate PR for vtproto.

tonistiigi (approved):

PR #210, opened 2024-09-28
copy: add support for non-octal mode setting

tonistiigi:
prerequisite for https://github.com/moby/buildkit/issues/1951

### DIFF
--- a/bench/go.mod
+++ b/bench/go.mod
@@ -1,6 +1,6 @@
 module github.com/tonistiigi/fsutil/bench
 
-go 1.20
+go 1.21
 
 require (
 	github.com/containerd/continuity v0.4.1
--- a/copy/copy_windows.go
+++ b/copy/copy_windows.go
@@ -37,6 +37,10 @@ func getFileSecurityInfo(name string) (*windows.SID, *windows.ACL, error) {
 }
 
 func (c *copier) copyFileInfo(fi os.FileInfo, src, name string) error {
+	if c.modeSet != nil {
+		return errors.Errorf("non-octal mode not supported on windows")
+	}
+
 	if err := os.Chmod(name, fi.Mode()); err != nil {
 		return errors.Wrapf(err, "failed to chmod %s", name)
 	}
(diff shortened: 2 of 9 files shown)


bench/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/tonistiigi/fsutil/bench
 
-go 1.20
+go 1.21
+
+toolchain go1.23.1
```
crazy-max, line 5: toolchain strikes again :sweat_smile:

copy/copy_windows.go
```diff
@@ -37,6 +37,10 @@ func getFileSecurityInfo(name string) (*windows.SID, *windows.ACL, error) {
 }
 
 func (c *copier) copyFileInfo(fi os.FileInfo, src, name string) error {
+	if c.modeSet != nil {
+		return errors.Errorf("non-octal mode not supported on windows")
+	}
```
crazy-max, line 42: Don't think a module exists for this but I think on Windows we could use `golang.org/x/sys/windows` to set corresponding Windows ACL permissions if we want that in the future (cc @gabriel-samfira)

crazy-max (approved):

bench/go.mod
tonistiigi: :rage1:

crazy-max (approved):

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.8.4
+	github.com/tonistiigi/dchapes-mode v0.0.0-20241001053921-ca0759fec205
```
thaJeztah, line 12: I see the original repository has a number of tagged versions already; should those be inherited? https://hg.sr.ht/~dchapes/mode/tags

tonistiigi: This is not the same source. I don't plan to make my own releases in the fork.

PR #214, opened 2024-10-22
fix correcting timestamps for created destination dir after copy

### LINKED ISSUE #207: copying a file with a tree path doesn't set timestamps correctly
This issue appeared while debugging issue with `buildkit` creating images with wrong timestamps. It is a custom LLB frontend which outputs an instruction like:

```
COPY --timestamp=<FIXED> /src/file /dst/path1/path2/file
```

In the output, the `/dst/path1/path2/file` has proper timestamp `<FIXED>`, while `/dst`, `/dst/path1`, `/dst/path1` all have wrong timestamps (current timestamps).

While debugging this, I got down to this library and put a unit-test (to `copy_test.go`) which fails:

```go
func TestCopyFileWithPathTimestamp(t *testing.T) {
	timestamp := time.Unix(0, 0)
	apply := fstest.Apply(
		fstest.CreateDir("/foo/", 0755),
		fstest.CreateFile("/foo/bar", []byte{}, 0644),
	)

	t1 := t.TempDir()
	t2 := t.TempDir()

	require.NoError(t, apply.Apply(t1))
	require.NoError(t, Copy(context.TODO(), t1, "/foo/bar", t2, "/test1/test2/test3/bar", WithCopyInfo(CopyInfo{
		CopyDirContents: true,
		Utime:           &timestamp,
	})))

	for _, s := range []string{"/test1/test2/test3/bar", "/test1/test2/test3", "/test1/test2", "/test1"} {
		stat, _ := os.Stat(filepath.Join(t2, s))
		require.Equal(t, timestamp, stat.ModTime(), "path: %s", s)
	}
}
```

This test shows exactly same as I observed with buildkit - the final file has a correct timestamp, while all intermediate directories have a current timestamp.

The reason is that `MkdirAll` recursively calls itself, and when it creates the parent `/path1`, the `/path1` timestamp will be set correctly with `Utimes`, but then `MkdirAll` will create `/path1/path2`, which will reset the timestamp of `/path1`, and so on.

The second issue is in the `Copy` function itself - when it creates the file, it doesn't revert the timestamp of the parent directory.

So I couldn't come up with an easy fix, so creating this issue to get some ideas how to fix it.

For me the most clean way would be to set the timestamp after all copy operations for all the created items to avoid dealing with parent-child file creation.

tonistiigi:
fixes #207

@smira

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -79,8 +80,10 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		if err != nil {
 			return err
 		}
-		if err := MkdirAll(ensureDstPath, 0755, ci.Chown, ci.Utime); err != nil {
+		if createdDirs, err := MkdirAll(ensureDstPath, 0755, ci.Chown, ci.Utime); err != nil {
 			return err
+		} else {
+			defer fixCreatedParentDirs(createdDirs, ci.Utime)
 		}
 	}
 
@@ -120,10 +123,11 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		if err != nil {
 			return err
 		}
-		dst, err := c.prepareTargetDir(srcFollowed, src, dst, ci.CopyDirContents)
+		dst, createdDirs, err := c.prepareTargetDir(srcFollowed, src, dst, ci.CopyDirContents)
 		if err != nil {
 			return err
 		}
+		defer fixCreatedParentDirs(createdDirs, ci.Utime)
 		if err := c.copy(ctx, srcFollowed, "", dst, false, patternmatcher.MatchInfo{}, patternmatcher.MatchInfo{}); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


smira (approved):
thank you for the fix!

copy/copy.go
```diff
@@ -79,8 +80,10 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 		if err != nil {
 			return err
 		}
-		if err := MkdirAll(ensureDstPath, 0755, ci.Chown, ci.Utime); err != nil {
+		if createdDirs, err := MkdirAll(ensureDstPath, 0755, ci.Chown, ci.Utime); err != nil {
 			return err
+		} else {
+			defer fixCreatedParentDirs(createdDirs, ci.Utime)
```
crazy-max, line 86: related to https://github.com/docker/buildx/pull/2745#discussion_r1810862520, maybe it could also apply ownership as well:

```suggestion
			defer fixCreatedParentDirs(createdDirs, ci.Chown, ci.Utime)
```

crazy-max, line 86: Related to https://github.com/docker/buildx/pull/2745#discussion_r1810862520, could we apply ownership as well?:

```suggestion
			defer fixCreatedParentDirs(createdDirs, ci.Chown, ci.Utime)
```

tonistiigi: The issue is that although target directories are created with requested timestamp, when files are written into them it changes the modified time. So the modified time needs to be reset again after the copy. This problem does not exist for chown.

There is another weird case in this code where target dirs are always created with default `755` mode, even if `--chmod` is set. Not sure if this is correct, but not directly part of this issue.

crazy-max (approved):

PR #215, opened 2024-11-01
Fixed build on OpenBSD

catap:
This fixes https://github.com/docker/buildx/issues/2772

### DIFF
--- a/copy/copy_openbsd.go
+++ b/copy/copy_openbsd.go
@@ -1,0 +1,38 @@
+//go:build openbsd
+// +build openbsd
+
+package fs
+
+import (
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+func copyFile(source, target string) error {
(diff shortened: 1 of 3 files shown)


tonistiigi (approved):
We should add this to CI as well. If the tests can't run, then at least check that it builds.

tonistiigi:
https://github.com/tonistiigi/fsutil/blob/master/docker-bake.hcl#L90

catap:
@tonistiigi seems that it requires a bit more work what just add it into this file. At least I see that dedicated instance with VM via vargand is required.

tonistiigi:
Vagrant would be needed to run the tests. To just check that code builds doesn't need it.

catap:
Here my attempt: https://github.com/catap/fsutil/tree/openbsd-ci

if you think that it’s enough I ready to open PR

catap:
Tuis, I had run it as: https://github.com/catap/fsutil/actions/runs/11635591513

PR #218, opened 2024-11-03
enable golangci-lint for supported platforms

crazy-max:


docker-bake.hcl
```diff
@@ -35,6 +39,16 @@ target "test-noroot" {
 target "lint" {
   dockerfile = "./hack/dockerfiles/lint.Dockerfile"
   output = ["type=cacheonly"]
+  platforms = GOLANGCI_LINT_MULTIPLATFORM != null ? [
+    "darwin/amd64",
```
tonistiigi, line 43: why is this list slightly different from the other one? We could also combine it using inherits (or attributes)

crazy-max: similar to buildkit / buildx, I recall we had issues with the linter for PowerPC and s390x architecture

crazy-max: switch to inherits and looks good: https://github.com/tonistiigi/fsutil/actions/runs/11671808580/job/32499036992#step:4:482

tonistiigi (approved):

PR #220, opened 2024-11-05
netbsd: fixes, build and test

crazy-max:
similar to https://github.com/tonistiigi/fsutil/pull/216

cc @akhramov @catap

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,7 @@ jobs:
       matrix:
         os:
           - freebsd
+          - netbsd
           - openbsd
     steps:
       -
--- a/copy/copy_otherbsd.go
+++ b/copy/copy_otherbsd.go
@@ -1,5 +1,5 @@
-//go:build openbsd
-// +build openbsd
+//go:build openbsd || netbsd
+// +build openbsd netbsd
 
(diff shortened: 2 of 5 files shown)


copy/copy_netbsd.go
akhramov, line 1: Is this identical to https://github.com/tonistiigi/fsutil/blob/8d32dbdd27d368aba5a636d014c1b9806f277f8b/copy/copy_freebsd.go? Probably we could just tweak build tags a bit and have the same file for both OSes?

crazy-max: Good catch yes we should

catap: I don't think so, see the last argument of unix.Mknod at the end of the file

crazy-max: @catap For both `freebsd` and `netbsd` it seems fine no?

crazy-max: Ah actually no, it's `openbsd` and `netbsd`

catap (approved):
I see that all test is passed on OpenBSD and NetBSD.

I haven't tested it locally, but I assume that it builds fine.

akhramov (approved):
lgtm

PR #226, opened 2025-01-14
fix: clear ModeIrregular on directories on Windows

fiam:
Since go1.23, directories with an unknown reparse tag will have the
irregular bit set. We do not know exactly which scenario causes this, but
there are many different reparse tags that can be set and go only handles
3 of them.

This causes issues with other parts of the code because in Unix a directory
cannot have this bit set.

Add a workaround to clear this bit for directories on Windows.

### DIFF
--- a/stat_windows.go
+++ b/stat_windows.go
@@ -15,3 +15,16 @@ func loadXattr(_ string, _ *types.Stat) error {
 
 func setUnixOpt(_ os.FileInfo, _ *types.Stat, _ string, _ map[uint64]string) {
 }
+
+func modeBits(fi os.FileInfo) uint32 {
+	mode := fi.Mode()
+	// Go 1.23 an above will set ModeIrregular for directories with reparse tags
+	// that it doesn't understand. Clear it to make the mode consistent with
+	// the expectations for ModeIrregular on Unix.
+	// See https://github.com/golang/go/blob/bd80d8956f3062d2b2bff2d7da6b879dfa909f12/src/os/types_windows.go#L227
+	// and https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4
+	if mode&os.ModeDir != 0 {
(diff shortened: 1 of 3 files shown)


fiam:
References:

- Go 1.23 implementation: https://github.com/golang/go/blob/bd80d8956f3062d2b2bff2d7da6b879dfa909f12/src/os/types_windows.go#L227
- Go pre1.23 implementation: https://github.com/golang/go/blob/bd80d8956f3062d2b2bff2d7da6b879dfa909f12/src/os/types_windows.go#L260
- Reparse tags: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4

crazy-max:
Seems your branch is not up to date with base one. Can you rebase?

stat_windows.go
```diff
@@ -15,3 +15,16 @@ func loadXattr(_ string, _ *types.Stat) error {
...
 }
+
+func modeBits(fi os.FileInfo) uint32 {
+	mode := fi.Mode()
+	// Go 1.23 an above will set ModeIrregular for directories with reparse tags
+	// that it doesn't understand. Clear it to make the mode consistent with
+	// the expectations for ModeIrregular on Unix.
+	// See https://github.com/golang/go/blob/bd80d8956f3062d2b2bff2d7da6b879dfa909f12/src/os/types_windows.go#L227
+	// and https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4
+	if mode&os.ModeDir != 0 {
```
jsternberg, line 26: Use `ModeType` here instead to match the behavior of pre go 1.23.

https://github.com/golang/go/blob/bd80d8956f3062d2b2bff2d7da6b879dfa909f12/src/os/types_windows.go#L260

jsternberg (changes requested):
Just one suggested change.

PR #233, opened 2025-04-08
Allow metadata-only receivers

tonistiigi:
This optional mode allows skipping transferring the file data if remote side just hopes to analyze how the local files look like.

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -1,0 +1,44 @@
+package fsutil
+
+import (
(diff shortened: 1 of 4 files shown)


crazy-max (approved):
Needs https://github.com/tonistiigi/fsutil/pull/234 for freebsd issue in ci

crazy-max (approved):

PR #238, opened 2025-06-05
fix: send: use platform-specific root path

profnandaa:
Using "/" was causing a silent bug later on at
`fs.go:121` that is expecting platform-specific
separators. See discussion at https://github.com/moby/buildkit/pull/6007

Fix this by using `\\` on Windows and `/` on unix.

### DIFF
--- a/send.go
+++ b/send.go
@@ -148,7 +148,8 @@ func (s *sender) sendFile(h *sendHandle) error {
 
 func (s *sender) walk(ctx context.Context) error {
 	var i uint32 = 0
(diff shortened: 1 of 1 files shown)


tonistiigi (approved):
I think we need to do a bit more thorough overview of the walkers and define more clearly which path should be used where. `proto.Stat.Path` should always be unix style while I think input path should be machine dependant. The callback path I think machine dependant but not 100% , if you look at the `filterFS` then it seems to be unix-style while if you look at `subDirFS` then it is machine-style. I think this one is correct anyway.

PR #242, opened 2025-10-01
stat: add statfs method to perform stat on fs.FS interface

jsternberg:
The primary purpose of this is to enable using the `types.Stat` type in
combination with `os.OpenRoot` which can return an `fs.FS` compatible
interface. This way, `StatFS` can be used on a chroot filesystem safely.

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -1,6 +1,7 @@
 package fsutil
 
 import (
+	gofs "io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


jsternberg:
This isn't going to work right now due to its reliance on go 1.25 and the projects that use this downstream aren't willing to set their minimum version to 1.25. I'm going to make another PR for just the mkstat portion.

PR #243, opened 2025-11-19
Fix Windows 'Access is denied' error on protected folders

rzlink:
Skip os.Lstat() for excluded paths when safe to prevent access errors on Windows protected system folders like 'System Volume Information' and 'WcSandboxState'.

Optimization logic:
- Check exclude patterns BEFORE calling os.Lstat()
- Skip Lstat if: path is excluded AND no include patterns AND no negations
- Still call Lstat when include patterns or negations exist (needed for matching)

Testing:
- Added testHookLstat field to CopyInfo for test verification
- 4 new Windows tests verify Lstat is skipped/called correctly:
  * TestCopyWithExcludeSkipsLstat
  * TestCopyMultipleExcludesSkipLstat
  * TestCopyWithIncludeAndExcludeMustCallLstat
  * TestCopyWithNegationPatternsMustCallLstat

### DIFF
--- a/copy/copy.go
+++ b/copy/copy.go
@@ -111,6 +111,7 @@ func Copy(ctx context.Context, srcRoot, src, dstRoot, dst string, opts ...Opt) e
 	if err != nil {
 		return err
 	}
+	c.testHookLstat = ci.testHookLstat
 	srcs := []string{src}
 
 	if ci.AllowWildcards {
@@ -206,6 +207,9 @@ type CopyInfo struct {
 	// replace any existing symlink or file)
 	AlwaysReplaceExistingDestPaths bool
 	ChangeFunc                     fsutil.ChangeFunc
+
+	// testHookLstat is called before each os.Lstat if non-nil (for testing only)
+	testHookLstat func(path string)
 }
 
 type Opt func(*CopyInfo)
@@ -272,6 +276,7 @@ type copier struct {
 	changefn                       fsutil.ChangeFunc
 	root                           string
 	alwaysReplaceExistingDestPaths bool
+	testHookLstat                  func(string)
 }
 
 type parentDir struct {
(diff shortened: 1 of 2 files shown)


rzlink:
@tonistiigi Could you review this PR? If the change seems risky for Linux, I can leave the Linux code untouched and make it Windows-only.

copy/copy.go
```diff
@@ -357,9 +358,42 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
...
+			return nil
 		}
 	}
 
+	if c.testHookLstat != nil {
+		c.testHookLstat(src)
+	}
+	fi, err := os.Lstat(src)
+	if err != nil {
+		return errors.Wrapf(err, "failed to stat %s", src)
```
thaJeztah, line 384: FWIW, `os.Lstat` (and many of the `os` functions) returns a `PathError`, which already includes the command executed and the path executed on;
https://github.com/golang/go/blob/2b62144069a130cc469f33009c0c392cc6de8810/src/os/stat_unix.go#L48


So wrapping it may make the error rather verbose (and duplicate the information); https://go.dev/play/p/pslhHfcWKDX

```console
failed to stat /no/such/file: lstat /no/such/file: no such file or directory
```

thaJeztah: Oh! I see that was already present, just moved (still worth looking at, perhaps for a follow-up 😅)

jsternberg (approved):

copy/copy.go
```diff
@@ -357,9 +358,42 @@ func (c *copier) copy(ctx context.Context, src, srcComponents, target string, ov
...
+	}
+
+	// After Lstat, if this item is excluded and is NOT a directory, skip it
+	if !include && !fi.IsDir() {
+		return nil
+	}
+
+	targetFi, err := os.Lstat(target)
+	if err != nil && !os.IsNotExist(err) {
+		return errors.Wrapf(err, "failed to stat %s", src)
```
tonistiigi: `target` in error message

copy/copy.go
rzlink: I'll leave the code as is and let @tonistiigi decide if he wants to address the duplicate information.

PR #247, opened 2026-05-21
copy: fix mode of created directories with unexpected umask

jsternberg:
Adds some code to `MkdirAll` to check the current umask for the process
and fix the permission mode of the file when the umask would interfere
with the creation of the directory.

### DIFF
--- a/copy/mkdir.go
+++ b/copy/mkdir.go
@@ -2,12 +2,19 @@ package fs
 
 import (
 	"os"
+	"path/filepath"
+	"sync"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


jsternberg:
Updated this with something that I think will work better.

The first time this makes a directory, it attempts to create a file and then observes the actual created mode of that file to determine what the umask is. Then, in the future, it just statically determines if the umask would be affected by the observed umask. I've also added some tests for this.

tonistiigi:
superceded https://github.com/tonistiigi/fsutil/pull/248

PR #253, opened 2026-06-03
ci: switch to pygolo/openbsd7 vagrant image

crazy-max:
relates to https://github.com/tonistiigi/fsutil/actions/runs/26913960532/job/79399003080#step:6:67

switch to pygolo/openbsd7 vagrant image like buildx: https://github.com/docker/buildx/blob/ff92e6e0bd6b406578f3188a77e54d07a8481c09/hack/Vagrantfile.openbsd#L5-L6

### DIFF
--- a/hack/Vagrantfile.openbsd
+++ b/hack/Vagrantfile.openbsd
@@ -2,7 +2,8 @@
 # vi: set ft=ruby :
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "generic/openbsd7"
+  config.vm.box = "pygolo/openbsd7"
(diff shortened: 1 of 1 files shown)


thaJeztah:
What does the `pygolo` stand for? Is that an individual, or something official?

crazy-max:
> What does the `pygolo` stand for? Is that an individual, or something official?

Yes seems an individual https://portal.cloud.hashicorp.com/vagrant/discover/pygolo/openbsd7. Only box working with libvirt afaik.

Generic one has not been updated for many years.

PR #261, opened 2026-06-08
ci: switch to bento/freebsd-14 vagrant box

crazy-max:
similar to https://github.com/moby/buildkit/pull/6833


crazy-max:
getting this one in to unblock ci

PR #268, opened 2026-07-06
remove legacy go build tags

crazy-max:


copy/stat_openbsd.go
```diff
@@ -1,5 +1,4 @@
 //go:build openbsd
-// +build openbsd
```
thaJeztah: Looks like `stat_openbsd` and `stat_sysv` are identical now, so probably can add `openbsd` as build tag to the other and remove this one.

copy/stat_sysv.go
thaJeztah, line 1: and possibly rename this one `stat_unixnobsd` (similar to `diskwriter_unixnobsd.go`) or `stat_unixnofreebsd` (which may be more accurate)

thaJeztah (approved):
LGTM, but left a suggestion (I was looking at these files when I saw you opened this PR 😂)

thaJeztah (approved):
LGTM

tonistiigi (approved):

